PR #30, opened 2021-02-10
Add RPC connections pool

wivern-co-uk:
- [x] Implement simultaneous connections to different validators' nodes
- [x] Remove arbitrary instantiation of web3 clients


wivern-co-uk:
@superbeefyboy, I assume there is no side effect when using different endpoints for reading operations and I didn't find write usages

superbeefyboy (approved):

PR #88, opened 2021-03-04
Add Julswap LP Pools

MirthFutures:
Adding BIFI-BNB, JULb-BNB, JulD-BNB and JulB-JULd


roman-monk:
we should reuse single /lps endpoint now, no need for '/jul/lps'

Issue #61, opened 2021-02-22
Unify all LP token prices under a single endpoint


sirbeefalot:
## Context 
We're trying to increase performance and decrease load of the Beefy API. We currently have different endpoints to track the value of LP tokens from different platforms. For example BUSD-BNB from pancakeswap and BUSD-BNB from streetswap, or kebab. 

## Problem
This causes redundancy by having to fetch multiple prices for assets that will be valued at pretty much the same price. It also means there's more files and lines of code that devs have to edit when adding pools. 
Devs have to add a bunch of handlers and this type of things will only get worse:
```
const oracleEndpoints = {
  'bakery-lp': () => fetchLP(endpoints.bakeryLp),
  bakery: () => fetchBakery(),
  'bdollar-lp': () => fetchLP(endpoints.bdollarLp),
  coingecko: ids => fetchCoingecko(ids),
  'jetfuel-lp': () => fetchLP(endpoints.jetfuelLp),
  'monster-lp': () => fetchLP(endpoints.monsterLp),
  'narwhal-lp': () => fetchLP(endpoints.narwhalLp),
  'nyanswop-lp': () => fetchLP(endpoints.nyanswopLp),
  pancake: () => fetchPancake(),
  'pancake-lp': () => fetchLP(endpoints.pancakeLp),
  'thugs-lp': () => fetchLP(endpoints.thugsLp),
  'kebab-lp': () => fetchLP(endpoints.kebabLp),
  'sponge-lp': () => fetchLP(endpoints.spongeLp),
};
```

## Solution
Just keeping all LP prices under a single endpoint. If we already have a BUSD-BNB LP price, we shouldn't need to add it again just because Kebab is different. We don't use the LP prices for anything mission critical so slight variations don't affect us. 
The goal would be to make it as simple as possible to add and maintain new LP prices.

wivern-co-uk:
On it!

sirbeefalot:
I removed the usage of the old endpoints on all the other repos. We should be able to remove the legacy endpoints from the API and close this issue.

PR #95, opened 2021-03-10
refactoring: remove old prices endpoints

wivern-co-uk:
Finishes #61

- [x] Remove `/:platform/price` endpoints
- [x] Remove `/:platform/lps` endpoints
- [x] Strict oracle/oracleId to catch 0 prices in development
- [x] Skip loading Fortube APY if API key undefined

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -34,9 +34,9 @@ yarn start
 
 After you start the API it can take a minute or two before you can fetch the APYs. We currently log `getApys()` to the console when all the data is available.
 
-**/:platform/price**: We have a few different endpoints to fetch price oracle information of different assets. We get most token prices from "/pancake/price" but there are some exceptions that come from "/bakery/price" for example. The intention is to eventually unify all token prices under the same endpoint.
+**/prices**: All token prices under the same endpoint (crosschain).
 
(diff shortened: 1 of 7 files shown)


roman-monk:
we're probably still using some endpoinds in cowllector
https://github.com/beefyfinance/beefy-cowllector/blob/e6cbd34152071f7bcc1bd7227b01d2e3ccbcd38b/utils/fetchPrice.js#L12

sirbeefalot:
> we're probably still using some endpoinds in cowllector
> https://github.com/beefyfinance/beefy-cowllector/blob/e6cbd34152071f7bcc1bd7227b01d2e3ccbcd38b/utils/fetchPrice.js#L12

Pretty sure I updated that file. The oracles aren't really used but the helpers were still there. I just removed everything, so this should be ok to merge.

superbeefyboy (approved):
Code seems good. We should merge it on the weekend or whenever we have some downtime between launches, just in case there is any issue / missing price.

wivern-co-uk:
Looks like old endpoints are still used in app. We would need to rename the one for pancake ang get rid of bakery one here: https://github.com/beefyfinance/beefy-app/blob/master/src/features/web3/fetchPrice.js

PR #116, opened 2021-03-25
Move daily earnings from BSCScan to Ethers Events

jpopxfile:
Daily earnings calculation now gives error due to bscscan rate limiting. Added sleep to not get rate limited.


jpopxfile:
As BSCScan does rate limit and can cause errors if too many requests are made, I have changed the implementation to use events instead.

superbeefyboy (approved):

PR #145, opened 2021-04-28
Add HFI and some Ape

MirthFutures:


src/api/stats/getAmmPrices.js
```diff
@@ -58,15 +58,17 @@ const marshPools = require('../../data/degens/marshLpPools.json');
...
 const INIT_DELAY = 60 * 1000;
 const REFRESH_INTERVAL = 10 * 60 * 1000;
 
 // FIXME: if this list grows too big we might hit the ratelimit on initialization everytime
 // Implement in case of emergency -> https://github.com/beefyfinance/beefy-api/issues/103
 const pools = [
+  ...hfiPools,
   ...comethPools,
-  ...popsiclePools,
+  //...popsiclePools,
```
roman-monk, line 71: @MirthFutures  this is needed?

MirthFutures: When I ran the api locally, it kept crashing if I didn't comment out popsicle.

roman-monk: should be not related with it, works on prod.
but i saw superb also commented it before.
very weird, its just another cake LP

PR #156, opened 2021-05-03
add up trading fee to pancakeswap

Kalesberg:


superbeefyboy (changes requested):
The api is currently not considering trading fees, so we are globally under-reporting yields to users. We only show the yield farmed, not the yield generated by trading fees.

This is a good direction  to start looking at the correct solution:
https://uniswap.org/docs/v2/protocol-overview/smart-contracts
https://uniswap.org/docs/v2/advanced-topics/understanding-returns/

PR #167, opened 2021-05-12
Added  ApeSwap Matic-BNB

JordanRaphael:
The lps are working correctly.

As I mentioned on #development, I was facing some issues with updating apys.

Tested on Windows 10 and Debian 10, the error is _apy initialization failed Error: Invalid JSON RPC response: ""_. (waited about an hour and every request had the same result)
I tried changing web3 version but same issues occured. I can test the api further if you could provide some info on it.

### DIFF
--- a/src/data/degens/apeLpPools.json
+++ b/src/data/degens/apeLpPools.json
@@ -1,5 +1,23 @@
 [
   {
+    "name": "banana-matic-bnb",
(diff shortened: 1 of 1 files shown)


JordanRaphael:
After wivern's tip I managed to run the API here is the APY

matic-bnb

sirbeefalot:
Great work! Will review and merge tomorrow this and the app PR first thing in the morning. Unless someone else does it before me.

PR #175, opened 2021-05-18
Reorganized stats folders for apy

JordanRaphael:
Using the recent mention of /tvl as motivation, I thought that the API folders are a bit discouraging to newcomers, hopefully it'll help with productivity as well.

I'd like your opinions at:
- There are multi-chain platforms e.g. complus that have multiple files under the same folder for each chain. Should we split them in to their corresponding chain folder?
- Should we do something similar for the AMM prices as well?

Tested on Debian


roman-monk:
> There are multi-chain platforms e.g. complus that have multiple files under the same folder for each chain. Should we split them in to their corresponding chain folder?
I think we can split yes, as they have different getAPY files on different chains

> Should we do something similar for the AMM prices as well? 
@wivern-co-uk do you have any thoughts on what would be the best way to reorganize amm prices? or do we need it at all?

JordanRaphael:
Accidentally I merged it with the tvl branch, I'm marking it as draft for now, will work on it soon to merge before there are too many changes in beefy-api.

PR #181, opened 2021-05-19
Add /tvl endpoint

JordanRaphael:
It's not 100% operational, still has some kind of _unhandled promise_ I managed to get 100% of vault tvl on BSC, Matic and Avax.

If anyone wants to experiment with it feel free. I'll try to patch it up tomorrow.


JordanRaphael:
It's working fine locally, I used cached /lps and /prices to get it out faster.
All chains are under /tvl endpoint, can be differentiated using each chain's id.

roman-monk:
getAvaxTvl, getBscTvl, ... all look same with different chainId and vault urls.
what do you think if we make a config and call 1 single `getChainTvl(chainVaults)`

```
const chains = [
  {
    chainId: 56,
    vaults: '...bsc_pools.js'
  },
  {
    chainId: 250,
    vaults: '...fantom_pools.js'
  },
];

chains.forEach(chain => promises.push(getChainTvl(chain)));
```



Web3 can be constructed from chainId as `const web3 = web3Factory(chainId)`

JordanRaphael:
> getAvaxTvl, getBscTvl, ... all look same with different chainId and vault urls.
> what do you think if we make a config and call 1 single `getChainTvl(chainVaults)`

Good call :) Code looks much cleaner now!

Removed the ```await sleep(100)``` since it was used for local testing. Tested without cached prices so looks good to me.

PR #191, opened 2021-05-25
Fix sushi apys

Moonster-BSC:
Unfortunately I don't have the root cause of why the APYs weren't adding up on refresh. Basically the issue seen is on periodic refresh of apys by the api, the trading aprs would show up as negative for sushi vaults. This code path was different than everywhere else (quick, cometh, polyzap). The code path was adapted to be very similar, but sushi requires a different query so it's slightly different. This was tested by leaving the api running indef and setting a breakpoint on the getSushiLPApys call, and making sure the subsequent trading apr is in the ballpark range


sirbeefalot (approved):
Tested it out. Seems to maintain the correct APY now, nice.

PR #193, opened 2021-05-26
Use previous full day data for trading APR

Moonster-BSC:
The latest DayPairData isn't complete, as its still being written to for that day, as pointed out by @roman-monk. This cause underestimation in volume. As a result, we use the data for the soonest day we have full daily data for, which is the entry between (now-2) days and (now-1) days


nloewen:
This is going to be great. I just watched the displayed rate for MATIC-USDC drop from over 0.6% daily to 0.09% as UTC rolled over to the 27th

PR #196, opened 2021-05-26
DFYN dual farm apy

Moonster-BSC:


roman-monk:
@Moonster-BSC it seems we can remove "blockchain-addressbook" and just use tokenSymbol as id

PR #201, opened 2021-05-28
[WIP] Get LP token info from Want token

Moonster-BSC:
We currently have to put the lp0 info and lp1 info for every pool in our json file. We could deduce this info instead of hardcoding it here, via the paircontract, and then perform a lookup in the addressbook. This PR demonstrates that concept. It is a WIP since it needs to run prior to any of the pool jsons being used, which is not the case, as they are sometimes directly imported in the getAPY calculation. It also needs some multicall optimization. Nevertheless, this concept can be reused to automate the pancake vault deployments, as @wivern-co-uk would like to do

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "apollo-link-http": "^1.5.17",
     "axios": "^0.20.0",
     "bignumber.js": "^9.0.1",
+    "blockchain-addressbook": "^0.1.15",
     "date-fns": "^2.21.3",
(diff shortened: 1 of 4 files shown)


wivern-co-uk:
I also want to automate this part, but my idea was to add a CLI command to add vaults with lp0 and lp1 to json files. Since this data is immutable it will be beneficial for overall app performance to keep having into json files.

PR #215, opened 2021-06-07
Goldenbull finance + addressbook refactors

Moonster-BSC:


src/api/stats/matic/getGoldenbullLpApys.js
```diff
@@ -0,0 +1,33 @@
...
+  },
+} = addressBook;
+const { quickLiquidityProviderFee } = require('./getQuickLpApys');
+const {
+  getScientificNotationFromTokenDecimals,
+} = require('../../../utils/getScientificNotationFromTokenDecimals');
+
+const getGoldenbullLpApys = async () =>
+  await getMasterChefApys({
+    masterchef: goldenbull.masterchef,
```
roman-monk, line 20: maybe we can keep masterChef address here? 
it seems not used anywhere else. 
ABI is here, LPs with all addresses are here, weird to update address-book package just for masterChef address, no?

src/api/stats/matic/getGoldenbullLpApys.js
```diff
@@ -0,0 +1,33 @@
...
+    masterchefAbi: MasterChefAbi,
+    tokenPerBlock: 'gBullPerBlock',
+    hasMultiplier: true,
+    pools: pools,
+    oracleId: GBULL.symbol,
+    oracle: 'tokens',
+    decimals: getScientificNotationFromTokenDecimals(GBULL.decimals),
```
roman-monk, line 27: nit: is it possible to have some extension function like `GBULL.eDecimals`, such a long name for just adding `1e` 😄

src/api/stats/matic/getGoldenbullLpApys.js
Moonster-BSC: Well i was thinking about refactoring the data as well, and abi could be moved to address-book as well. There's checksum benefit as well in address-book. We could open up the discussion to others

src/api/stats/matic/getGoldenbullLpApys.js
Moonster-BSC: I can't quickly figure out how to maintain the const typing while adding the extension. I'll just shorten the name for now

PR #216, opened 2021-06-07
Feat/apy endpoint migration

Tumble17:
PR creates a new endpoint in the API at /apy/breakdown. New endpoint was requested from Weso.

First PR brings in SpookyLPs, pauses to check that we like the format and then future PR(s) will iterate through LPs that are compatible with the new APY endpoint and have all the parameters required to expose these extra values.

Existing /apy endpoint is left as it was and has been referred to as the 'legacy' APY endpoint in the code.

In order to maintain both endpoints until we are able to migrate to the new /apy/breakdown endpoint fully, if we would like to, we need to maintain both object structures. We create an object that holds apy: and apyBreakdowns: objects and append to these in the same way we did before. At each level, we then check for the existence of these structures and continue breaking them out from the platform level LPs all the way up to getApys.

For this PR with spooky, this breakdown happens in getSpookyLpApys -> index.js (Fantom level) -> getApys -> index.js (Stats Level).

At the LP level, this PR aims to expose as much of the totalAPY logic as feels appropriate. This has included:

simpleApy - yearlyRewardsInUsd / totalStakedInUsd
compoundingsPerYear - BASE_HYP pulled from constants
beefyPerformanceFee - 0.045 or (1-0.955), which has also been refactored to a variable
vaultApy - simpleApy compounded
lpFee - Liquidity provider fee specific to platform
tradingApy - Pulled from Apollo (Subgraph)
totalApy - Full compounded APY using full calculation. This value is the same as the only value in the legacy /apy endpoint

Where the values are not available and therefore cannot be exposed in both the new and legacy formats, the legacy format is brought forward and mapped into the new endpoint. This means that you will see the totalApy for vaults where the breakdown does not exist.

### DIFF
--- a/src/api/stats/getApys.js
+++ b/src/api/stats/getApys.js
@@ -40,8 +40,17 @@ const updateApys = async () => {
 
       // Set default APY values
       let mappedApyValues = result.value;
-      let mappedApyBreakdownValues = result.value;
+      let mappedApyBreakdownValues = {};
 
+      // Loop through key values and move default breakdown format
+      // To require totalApy key
+      for (const [key, value] of Object.entries(result.value)) {
+        mappedApyBreakdownValues[key] = {
+          totalApy: value,
+        };
+      }
+
+      // Break out to apy and breakdowns if possible
       let hasApyBreakdowns = 'apyBreakdowns' in result.value;
       if (hasApyBreakdowns) {
         mappedApyValues = result.value.apys;
--- a/src/api/stats/getApys.js
+++ b/src/api/stats/getApys.js
@@ -3,13 +3,14 @@ const getBifiMaxiApy = require('./beefy/getBifiMaxiApy');
 const { getAvaxApys } = require('./avax');
 const { getMaticApys } = require('./matic');
 const { getHecoApys } = require('./heco');
-const { getFantomApys } = require('./fantom');
+const { getFantomApys, getFantomApyBreakdowns } = require('./fantom');
 const { getBSCApys } = require('./bsc');
 
 const INIT_DELAY = 30 * 1000;
 const REFRESH_INTERVAL = 15 * 60 * 1000;
 
 let apys = {};
+let apyBreakdowns = {};
 
 const getApys = () => {
   return apys;
@@ -33,6 +34,21 @@ const updateApys = async () => {
         console.warn('getApys error', result.reason);
         continue;
       }
+
+      // // Set default APY values
+      // let mappedApyValues = result.value
+      // let mappedApyBreakdownValues = result.value
+
+      // let hasApyBreakdowns = "apyBreakdowns" in result.value
+      // if (hasApyBreakdowns) {
+      //   mappedApyValues = result.value.apys
+      //   mappedApyBreakdownValues = result.value.apyBreakdowns
+      // }
+
+      // apys = { ...apys, ...mappedApyValues };
+
+      // apyBreakdowns = { ...apyBreakdowns, mappedApyBreakdownValues };
+
       apys = { ...apys, ...result.value };
     }
 
@@ -46,4 +62,4 @@ const updateApys = async () => {
 
 setTimeout(updateApys, INIT_DELAY);
 
-module.exports = getApys;
+module.exports = { getApys };
--- a/src/api/stats/getApys.js
+++ b/src/api/stats/getApys.js
@@ -3,7 +3,7 @@ const getBifiMaxiApy = require('./beefy/getBifiMaxiApy');
 const { getAvaxApys } = require('./avax');
 const { getMaticApys } = require('./matic');
 const { getHecoApys } = require('./heco');
-const { getFantomApys, getFantomApyBreakdowns } = require('./fantom');
+const { getFantomApys } = require('./fantom');
 const { getBSCApys } = require('./bsc');
 
 const INIT_DELAY = 30 * 1000;
@@ -13,7 +13,10 @@ let apys = {};
 let apyBreakdowns = {};
 
 const getApys = () => {
-  return apys;
+  return {
+    apys,
+    apyBreakdowns,
+  };
 };
 
 const updateApys = async () => {
@@ -35,21 +38,19 @@ const updateApys = async () => {
         continue;
       }
 
-      // // Set default APY values
-      // let mappedApyValues = result.value
-      // let mappedApyBreakdownValues = result.value
+      // Set default APY values
+      let mappedApyValues = result.value;
+      let mappedApyBreakdownValues = result.value;
 
-      // let hasApyBreakdowns = "apyBreakdowns" in result.value
-      // if (hasApyBreakdowns) {
-      //   mappedApyValues = result.value.apys
-      //   mappedApyBreakdownValues = result.value.apyBreakdowns
-      // }
+      let hasApyBreakdowns = 'apyBreakdowns' in result.value;
+      if (hasApyBreakdowns) {
+        mappedApyValues = result.value.apys;
+        mappedApyBreakdownValues = result.value.apyBreakdowns;
+      }
 
-      // apys = { ...apys, ...mappedApyValues };
+      apys = { ...apys, ...mappedApyValues };
 
-      // apyBreakdowns = { ...apyBreakdowns, mappedApyBreakdownValues };
-
-      apys = { ...apys, ...result.value };
+      apyBreakdowns = { ...apyBreakdowns, ...mappedApyBreakdownValues };
     }
 
     console.log('> updated apys');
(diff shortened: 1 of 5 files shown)


src/api/stats/getApys.js
```diff
@@ -33,7 +37,20 @@ const updateApys = async () => {
...
+
+      let hasApyBreakdowns = 'apyBreakdowns' in result.value;
+      if (hasApyBreakdowns) {
+        mappedApyValues = result.value.apys;
+        mappedApyBreakdownValues = result.value.apyBreakdowns;
+      }
+
+      apys = { ...apys, ...mappedApyValues };
+
+      apyBreakdowns = { ...apyBreakdowns, ...mappedApyBreakdownValues };
```
Moonster-BSC, line 53: Do we want to add a default value to the apyBreakdowns? I think this line should be in the `if (hasApyBreakdowns)` statement, that way if the UI doesn't find a breakdown from the breakdown route it can default to the legacy apy endpoint values

Moonster-BSC (changes requested):

src/api/stats/getApys.js
Tumble17: I hope that’s in line 43, defaults previous to the has command. Does that cover what you meant?

Moonster-BSC: So what i mean is we have two final objects, apys and apyBreakdowns. If `apyBreakdowns` is not in result.value, then mappedApyBreakdownValues will look like the same object as the current apy object (= result.value). Then we'll have a mix of formats in the new endpoint, i.e.
```json
{
  "bifi-maxi": 0.13,
  "polycat-pzap-quick": {
        "simpleApy": ".13",
        "compoundingsPerYear": 1000, 
        // ...
      }
  }
}
```

Basically i think the formats shouldn't be mixed for the endpoint. If the breakdown is missing, it should be missing from the new endpoint.

Moonster-BSC: Its much easier for the app to check for a missing property (`if (key in  map)`) than to try to differentiate between two types, especially since we're using javascript instead of typescript

Tumble17: Right, understand what you’re saying. Thanks for explaining again. 

This was by design, so that logic would be done api side rather than app side. 

It effectively removes the need for ‘if not exists in A, go fetch from B’ so reduces calls. I am comfortable with and desired the hybrid mapping personally. The endpoint represents vaults APY as broken down as possible, which in the case of many is currently not at all.

Are your views on this as strong as blocking until changed?

Moonster-BSC: So the apys for all vaults are already located in the same endpoint. We fetch the apys for all vaults periodically in the app, with this one api call. With this new endpoint, we'd be adding the extra fetch call to the periodic fetch, now making 2 calls per interval. So we won't be saving from making extra calls, since the calls aren't individualized to the vaults.

I don't like having mixed formats in an api as it doesn't follow good rest practices. By having it be the same format, its easily described by a OpenAPI/Swagger spec or typescript interface, and makes it easy for others to consume. Think about the interface that entails:

```ts
interface BreakdownApy {
  [poolName: string]: number | { simpleApy: number, compoundingsPerYear: number, /*...*/ }
}
```

Anyone who wants to consume this api will need some custom logic to differentiate between the two formats.

This scenario described in this answer is indeed a contrived. We don't use typescript, and the api is mainly internal to beefy. But it still makes it easier to consume for our app

Tumble17: Thanks for going over how the app will consume this. I’m now in full agreement with you that this will be better as just the values that it has been possible to break down the APYs for. Will change and come back. Thanks for leaning into conversation.

ReflectiveChimp: I think you guys might just be misunderstanding each other?

@Moonster-BSC Sounds like you want the data returned like this?
```json
{
  "venus-wbnb": {
    "totalApy": 0.06983201937728578,
  },
  "boo-bifi-ftm": {
    "simpleApy": 1.1845128790579593,
    "compoundingsPerYear": 2190,
    "beefyPerformanceFee": 0.045,
    "vaultApy": 2.0984988299993503,
    "lpFee": 0.002,
    "tradingApr": 0.052152730701897375,
    "totalApy": 2.2643800119210837
  },
}
```

Instead the current output of:
```json
{
  "venus-wbnb": 0.06983201937728578,
  "boo-bifi-ftm": {
    "simpleApy": 1.1845128790579593,
    "compoundingsPerYear": 2190,
    "beefyPerformanceFee": 0.045,
    "vaultApy": 2.0984988299993503,
    "lpFee": 0.002,
    "tradingApr": 0.052152730701897375,
    "totalApy": 2.2643800119210837
  },
}
```

Moonster-BSC: This can work as well. As the "missing breakdown" format is a subset of the breakdown format. In this case the check is similar to a missing check. if (tradingApy in pool).

Moonster-BSC: This makes the interface: 
```ts
interface BreakdownApy {
  [poolName: string]: { 
        simpleApy: number,
        tradingApr?: number, // note the nullable ? 
        /*...*/ }
}
```

Moonster-BSC: But actually, now totalApy is kinda like how it is right now: we're not sure if it has trading in it or not from the number itself, we have to inspect the rest of the objects properties

Moonster-BSC: So i'd rather have the property in the missing breakdown be something like `"farmApy": 0.13, where farmApy is compounded farmAPR (simpleAPY is farmAPR, but honestly should be renamed). Like this:

```ts
{
  "venus-wbnb": {
     "farmApr": 0.05
    "farmApy": 0.06983201937728578,
  },
  "boo-bifi-ftm": {
    "farmApr": 1.12, // don't think this number is necessary tho
    "farmApy": 1.1845128790579593,
    "compoundingsPerYear": 2190,
    "beefyPerformanceFee": 0.045,
    "vaultApy": 2.0984988299993503,
    "lpFee": 0.002,
    "tradingApr": 0.052152730701897375,
    "totalApy": 2.2643800119210837
  },
}
```

Tumble17: I hadn’t actually thought of the approach of aligning the schema from your response, so thanks for that. I like this approach and also like your differentiation of the non fee totalApy -> vaultApy and the totalApy when full calculation is available. I will make changes to this now

Moonster-BSC: Just to clarify, are you going with @ReflectiveChimp 's suggestion, and what we have in my last comment? I think thats the best of both worlds: we get the functionality you were looking for (don't need to rely on fallback data from another endpoint), and we maintain a clean format

ReflectiveChimp: `totalApy` is never going to be the real total for **every single vault** unless they **all** have endpoints for trading fees.
So its really "knownTotalApy". 

Given this, probably okay to set `totalApy` = `vaultApy` for legacy data; making `totalApy` required -> i.e. the value consumers of API can rely on to exist and display with the rest being optional breakdown data?

Not a huge deal but otherwise for the 'Overview APY' we'd have to `totalApy in apys ? apys.totalApy : apys.vaultApy`

Moonster-BSC: My opinion isn't is strong on totalApy vs vaultApy. I think your argument for going with totalApy is reasonable. I was more concerned about the first issue regarding data format

Moonster-BSC: Thanks @ReflectiveChimp for the formatting suggestion

Tumble17: Very good, thanks both. For total clarity, that leads us to:

{
  "venus-wbnb": {
    "totalApy": 0.06983201937728578,
  },
  "boo-bifi-ftm": {
    "vaultApr": 1.1845128790579593, // rename from simpleApy
    "compoundingsPerYear": 2190,
    "beefyPerformanceFee": 0.045,
    "vaultApy": 2.0984988299993503,
    "lpFee": 0.002,
    "tradingApr": 0.052152730701897375,
    "totalApy": 2.2643800119210837
  },
}

Just also calling out that without going through other lp files, the vaultApr that appeared in the previous comment won't appear as that's not currently passed through. We will have totalApy in both broken down and non broken down cases and know that some vaults don't have the ability to be broken down yet. Will rename simpleApy as vaultApr as you suggested. Completely agree on the aligned schema with nullable now we've been through this discussion, we should be strong on that.

Tumble17:
@Moonster-BSC and @ReflectiveChimp updates made as we all discussed.

Output looks like the following now:

/apy (same as before):

```
"polycat-sushi-fish-matic": 129.6185042864676,
"boo-bifi-ftm": 2.258432458998008,
```

/apy/breakdown:
```
"polycat-sushi-fish-matic": {
        "totalApy": 129.6185042864676
},
"boo-bifi-ftm": {
        "vaultApr": 1.1826023473229887,
        "compoundingsPerYear": 2190,
        "beefyPerformanceFee": 0.045,
        "vaultApy": 2.0928535050965476,
        "lpFee": 0.002,
        "tradingApr": 0.052152730701897375,
        "totalApy": 2.258432458998008
},
```

PR #232, opened 2021-06-14
Enable TypeScript

BrknRobot:


superbeefyboy (commented):
Looks good. Do we need to migrate the api to TypeScript for any particular reason?

PR #247, opened 2021-06-20
If some yield data is missing, return what we've got

BrknRobot:


ReflectiveChimp:
These changes seem to be based on a old master. 

e.g. `vaultApr` was updated to have the performance fee taken in to account at https://github.com/beefyfinance/beefy-api/commit/d093c1f6e0daa36e1a05575491a6c4beb372db5d

Tumble17:
Nice work refactoring the split APY and abstracting that code block

src/api/stats/common/getApyBreakdown.ts
```diff
@@ -0,0 +1,51 @@
+import BigNumber from "bignumber.js";
+
+import getFarmWithTradingFeesApy from '../../../utils/getFarmWithTradingFeesApy';
+import { compound } from '../../../utils/compound';
+
+import { BASE_HPY } from '../../../constants';
+
+const beefyPerformanceFee = 0.045;
```
Tumble17, line 8: It was discussed in development https://discord.com/channels/755231190134554696/758368074968858645/855756241715462154  that the performance fee can change, so if that’s true in practice can we parameterise that please

Tumble17 (changes requested):
Performance fee change discussion from development

PR #281, opened 2021-07-08
Add Mai Finance

BrknRobot:


src/api/stats/matic/index.js
```diff
@@ -40,6 +41,7 @@ const getApys = [
   getDfynLpApys,
   getDfynDualFarmLpApys,
   getBoneSwapApys,
+  getMaiApys,
```
roman-monk, line 44: for lps we need to add them on top of the list, its read from bottom to top

PR #282, opened 2021-07-09
Add titan-busd and one-wbnb, rename BigONE

wivern-co-uk:
1. Add titan-busd and one-wbnb
2. Rename BigONE to avoid price conflict with Harmony's ONE
 
Requires coordinated deploy with https://github.com/beefyfinance/beefy-app/pull/625


roman-monk:
lets keep original id `cakev2-one-busd` unchanged ?
might have problem with tvl reported by cowllector, or some third-party service like yieldwatch

PR #306, opened 2021-07-23
add cakev2-c98-wbnb

0xww:


wivern-co-uk:
@0xww in order to enable zap afterwards you would need to add C98 to address-book:

packages/address-book/address-book/bsc/tokens/tokens.ts

0xww:
PR on #308

PR #308, opened 2021-07-23
addressbook/tokens/add coin98

0xww:


MirthFutures:
The checksum test failed on this. Can you checksum these addresses?

0xww:
@MirthFutures address was all in lowercase, now it is fixed

PR #316, opened 2021-07-27
Added autofarm addresses

0xww:


packages/address-book/address-book/bsc/platforms/auto.ts
```diff
@@ -0,0 +1,4 @@
+export const auto = {
+  masterchef: '0x0895196562C7868C5Be92459FaE7f877ED450452',
+  router: '0x05fF2B0DB69458A0750badebc4f9e13aDd608C7F', // can use pancake router
```
Moonster-BSC, line 3: I would remove v1 pancakeswap router. Anyways, its in pancake.ts

roman-monk:
i think we wont use their masterchef anymore 😄

PR #317, opened 2021-07-27
Added bakery addresses

0xww:


packages/address-book/address-book/bsc/platforms/bakery.ts
```diff
@@ -0,0 +1,4 @@
+export const bakery = {
+  masterape: '0x20eC291bB8459b6145317E7126532CE7EcE5056f',
```
Moonster-BSC, line 2: Why is this called masterape? Contract is called BakeryMaster. Is it cause its a fork of apeswap masterchef and you wanted to make a distinction between that fork and the typical masterchef?

PR #329, opened 2021-07-30
Added Wault DEP-BNB

JordanRaphael:
DEP-BNB APY: 12.5k%


MirthFutures:
Does DEP need to be added to tokens?

JordanRaphael:
@MirthFutures sorry for being late, just added it :)

PR #354, opened 2021-08-12
Add getBlockTime helper

sircryptomaxi:
Averages the time for blocks to be mined over the past 100 blocks. APYs will be more accurate if the block times are dynamic.

### DIFF
--- a/src/utils/getBlockTime.js
+++ b/src/utils/getBlockTime.js
@@ -1,0 +1,30 @@
+const { web3Factory } = require('./web3');
+
+const fastestChainBlockTimeInMilliseconds = 3000;
+const blockPeriod = 100;
+
+let cache = {};
+const getBlockTime = async chainId => {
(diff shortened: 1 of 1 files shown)


roman-monk:
i think i saw the case when getBlockNumber returns the number which is not yet available via getBlock(), because of RPC is not synced yet probably

sircryptomaxi:
> i think i saw the case when getBlockNumber returns the number which is not yet available via getBlock(), because of RPC is not synced yet probably

Shall I kick back `currentBlock` by some number to make sure that the block information would be available to the RPC?

ReflectiveChimp:
Instead of getting latest block number via `getBlockNumber` and then calling `getBlock` with that number, try just calling `getBlock('latest')`, and then going back X blocks from there.

I believe `getBlock('latest')` should always return the latest block available to the node (which may or may not match `getBlockNumber`)

ReflectiveChimp:
Also not sure you need to invalidate the cache every 3 seconds, I'm sure something like 5 minutes or so would be fine.

PR #363, opened 2021-08-23
Fix WETH entry

dcFX0:
WETH was mislabeled as WMATIC_DFYN
based upon the contractg address


roman-monk:
thats really DFYN's version of Matic 🤷‍♂️
WETH is this one https://polygonscan.com/address/0x7ceb23fd6bc0add59e62ac25578270cff1b9f619

PR #421, opened 2021-10-06
Update tokens.ts

mjoaris:
One note: I particularly like my description of Sushi over what I had found in the other chains. Might be worth cross-copying?


MirthFutures:
Can you add a , after the website for USDCe? It caused the build to fail.

PR #441, opened 2021-10-17
Feature/jarvis

Qkyrie:
- Added Jarvis Reward Token (JRT)
- Added JRT-ETH pair + Sushi rewardPool


MirthFutures:
Seems like Reward Pool is the MiniChef address, I don't believe this is correct. Does this pair give extra rewards?

Qkyrie:
It does: 
image

PR #501, opened 2021-11-23
Add Bison

sircryptomaxi:


src/api/stats/common/getRewardPoolApys.ts
```diff
@@ -72,12 +75,15 @@ const getPoolsData = async params => {
...
   const rewardRateCalls = [];
   params.pools.forEach(pool => {
-    const rewardPool = new web3.eth.Contract(IRewardPool, pool.rewardPool);
+    const abi = params.tokenPerBlock ? chefAbi(params.tokenPerBlock) : IRewardPool;
+    const tokenPerBlock = params.tokenPerBlock ?? 'rewardRate';
+    const tokenContract = new params.web3.eth.Contract(ERC20, pool.address);
+    const rewardPool = new params.web3.eth.Contract(abi, pool.rewardPool);
     balanceCalls.push({
-      balance: rewardPool.methods.totalSupply(),
+      balance: tokenContract.methods.balanceOf(pool.rewardPool),
```
roman-monk, line 83: this is incorrect, please revert.
i would suggest to revert whole file actually.
what you need is very likely master chef apy, or just custom degen/apy unless its super specific, but i think commonMasterChef will cover it without any internal changes.

sircryptomaxi: Changes to common files reverted. Added custom apy as it's a reward pool structure but the common files don't cover the functions.

roman-monk: its really not rewardPool, rewardPool is always per seconds and without allocPoints.
this is almost perfect master chef, just `totalAllocPoint` is private

PR #513, opened 2021-11-28
quick-matic-wcro

Qkyrie:
- added wcro (yeah, it's called like that even on polygon)
- added quick-matic-wcro


roman-monk:
small conflict, can try to add lp between existing pairs next time if there is another pr with same file changes

Qkyrie:
conflict fixed, pulled master but not sure why the other changes are suddenly there. They are non-functional though.

PR #524, opened 2021-12-08
feat: added /vaults/stake for BSC

Tumble17:
Creates a new endpoint, `vaults/stake` to expose the staked pools files JSON.

This PR only includes BSC pools on request from Pacoca and prioritises that requirment, but a further small refactor to include all chains would be very simple.

Schema:

``` typescript
interface stakeVaults {
    [
         {
             id: string,
             name: string,
             logo: string,
             token: string,
             tokenDecimals: number,
             tokenAddress: string,
             tokenOracle: string,
             tokenOracleId: string,
             earnedToken: string,
             earnedTokenDecimals: number,
             earnedTokenAddress: string,
             earnContractAddress: string,
             earnContractAbi: object,
             earnedOracle: string,
             earnedOracleId: string,
             partnership: boolean,
             status: string,
             fixedStatus: boolean,
             partners: object,
             chain: string,
         }
        ...
    ]
}
```

### DIFF
--- a/src/utils/getStakeVaults.js
+++ b/src/utils/getStakeVaults.js
@@ -1,0 +1,33 @@
+const axios = require('axios');
+const fs = require('fs');
+const { sleep } = require('./time');
+const { ABI_ENDPOINT } = require('../constants');
+
+const getStakeVaults = async vaultsEndpoint => {
+  try {
+    const response = await axios.get(vaultsEndpoint);
+    const data = response.data;
+    fs.writeFile('src/utils/pools/stake/bsc_stake.js', data, function (err) {
+      if (err) return console.log(err);
+      // console.log('> stake pools sourced into src/utils/pools/stake/bsc_stake.js');
+    });
+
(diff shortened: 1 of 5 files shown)


roman-monk:
why dont they just read `https://raw.githubusercontent.com/beefyfinance/beefy-app/master/src/features/configure/stake/bsc_stake.js`?

src/utils/getStakeVaults.js
```diff
@@ -0,0 +1,33 @@
+const axios = require('axios');
+const fs = require('fs');
+const { sleep } = require('./time');
+const { ABI_ENDPOINT } = require('../constants');
+
+const getStakeVaults = async vaultsEndpoint => {
+  try {
+    const response = await axios.get(vaultsEndpoint);
+    const data = response.data;
+    fs.writeFile('src/utils/pools/stake/bsc_stake.js', data, function (err) {
```
roman-monk, line 10: this seems a bit weird, i think we should not write src files in runtime.

PR #532, opened 2021-12-16
add eth-dai

Marth007:
Add eth-dai to sushiswap moonriver


roman-monk:
checksum failed https://github.com/beefyfinance/beefy-api/runs/4554309854?check_suite_focus=true

PR #541, opened 2021-12-21
Fix multicall dependency for strategy & lastHarvest fetch

sircryptomaxi:
Use the common multicall address for fetching strategy address and last harvest info instead of a custom multicall. Will make it easier to deploy to new chains without additional infra overhead.

Link T-167.


sircryptomaxi:
Link T-167

PR #549, opened 2021-12-26
Add Beethovenx Variations of a Theme USDC

sircryptomaxi:
asUSDC has no liquidity outside of this pool, but is closely pegged to USDC. asUSDC price can't easily be independently calculated so line 17 & 18 is meant to fetch the price for USDC using 18 decimals.

### DIFF
--- a/packages/address-book/address-book/fantom/tokens/tokens.ts
+++ b/packages/address-book/address-book/fantom/tokens/tokens.ts
@@ -904,6 +904,17 @@ const _tokens = {
     description:
       "Terra's native token, LUNA, is used to stabilize the price of the protocol's stablecoins. LUNA holders are also able to submit and vote on governance proposals, giving it the functionality of a governance token.",
   },
+  asUSDC: {
+    name: 'USDC Solana',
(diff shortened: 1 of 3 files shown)


packages/address-book/address-book/fantom/tokens/tokens.ts
```diff
@@ -904,6 +904,17 @@ const _tokens = {
     description:
       "Terra's native token, LUNA, is used to stabilize the price of the protocol's stablecoins. LUNA holders are also able to submit and vote on governance proposals, giving it the functionality of a governance token.",
   },
+  asUSDC: {
+    name: 'USDC Solana',
+    address: '0xb5E4D17FFD9D0DCE46D290750dad5F9437B5A16B',
+    symbol: 'USDC',
```
roman-monk, line 910: this should be asUSDC?

roman-monk:
can hardcode asUSDC price to 1, same as USDC is hardcoded in `knownPrices`

packages/address-book/address-book/fantom/tokens/tokens.ts
sircryptomaxi: Yes, it's corrected in the latest commit.

PR #556, opened 2021-12-30
feat: use ethers and Promises All to consult data

0xww:
Please, someone review and test this code before.

In local computer, only consume 40 MB.

### DIFF
--- a/src/api/stats/getMultichainVaults.js
+++ b/src/api/stats/getMultichainVaults.js
@@ -1,8 +1,9 @@
+const { ethers } = require('ethers');
 const getVaults = require('../../utils/getVaults.js');
-const { getStrategies } = require('../../utils/getStrategies.js');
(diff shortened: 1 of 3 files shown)


roman-monk:
whole app uses ~400MB on heroku with getVaults disabled.
i'm wondering what it was before the changes, can it be related with eth-multicall.

anyways i'm willing to keep current implementation, there is no leak so we can bump RAM, i dont think there is a big memory savings, is there?
multicall solution is better, as with promises we make 2 requests per each vault (thats 2500 requests), so some RPC might fail especially when it overlaps with prices or apys requests.
there is ethers-multicall libs but i dont think it can be significantly better.

some optimization could be to combine strategies and lastHarvests in 1 multicall, as this functions used only in getVaults.

0xww:
you are right. Multicall resolve request faster:
- now I testing current prod version (with getVaults enable) and refactor version and both seems to use same memory space.

PR #569, opened 2022-01-11
Strategy: Fix UST-MATIC LP

shadoweddev:
Fixed LP strategy by adjusting LP route from UST to UST wormhole and updating the address book to contain this address.

### DIFF
--- a/packages/address-book/address-book/polygon/tokens/tokens.ts
+++ b/packages/address-book/address-book/polygon/tokens/tokens.ts
@@ -27,6 +27,17 @@ const MAI = {
 } as const;
 
 const _tokens = {
+  USTw: {
+    name: 'TerraUSD (Wormhole)',
+    symbol: 'WORMHOLE',
+    address: '0xE6469Ba6D2fD6130788E0eA9C0a0515900563b59',
+    chainId: 137,
+    decimals: 18,
+    logoURI: 'https://s2.coinmarketcap.com/static/img/coins/64x64/15610.png',
+    website: 'https://coinmarketcap.com/currencies/terrausd-wormhole/',
+    description:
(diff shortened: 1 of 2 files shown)


MirthFutures:
Can you change the name to wormhole as well? 
    "name": "quick-UST-MATIC",

shadoweddev:
> Can you change the name to wormhole as well? "name": "quick-UST-MATIC",

I assume I should rename id: 'quick-ust-matic' in the app to mirror this change. what are your thoughts on renaming more fields in the app [PR](https://github.com/beefyfinance/beefy-app/pull/923/commits/df8e423226383c10817813eae1e4ee8f4ab66fc6)? wormhole is a denomination of UST so would it make sense to keep the token and name as "UST-MATIC LP" and keep the tokenEarned as 'mooQuickUST-MATIC'?

roman-monk:
why the token is `WORMHOLE`? 
its unrelated to UST project, we cant call all bridged tokens WORMHOLE.
symbol is "UST", so "USTw" makes more sense.

can keep original symbol, just change the token id
```
USTw : {
   symbol: 'UST',
}
```

roman-monk (changes requested):
also its 6 decimals.
please run API locally to test price and APY matches farm

roman-monk:
need to fix token description, its about wormhole bridge project, but the token is UST.
also lets use oracleId: 'USTw', just found they have different decimals so we cant reuse it, it will break calculations

shadoweddev:
I overlooked the description + will change. 
quick question: [USTw](https://polygonscan.com/address/0xE6469Ba6D2fD6130788E0eA9C0a0515900563b59#code) is not an ERC20. When I change oracleId to map to USTw address, calculation breaks. Is this due to USTw being a different standard than API calculation expected?
calculation works fine now w/oracle id UST even with decimal issue...

PR #571, opened 2022-01-12
Add QuickSwap KIRO-MATIC and LMT-ETH

pumpingGhost:


pumpingGhost:
Removed the LMT-ETH farm.
Moved KIRO-MATIC to the dual farm section and added the rewardB property.
Ready for review

PR #584, opened 2022-01-20
fix USTw

shadoweddev:
issue: USTw-MATIC deposit show 50% of value, this is due to an issue pulling USTw price.
solution: change symbol to UST. now USTw references UST price, which already exists in API.

### DIFF
--- a/packages/address-book/address-book/polygon/tokens/tokens.ts
+++ b/packages/address-book/address-book/polygon/tokens/tokens.ts
@@ -29,14 +29,14 @@ const MAI = {
 const _tokens = {
   USTw: {
     name: 'TerraUSD (Wormhole)',
-    symbol: 'USTw',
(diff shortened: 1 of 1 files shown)


MirthFutures:
This is wrong. The addresbook has nothing to do with the asset pricing in the API. This change with make the app break as two tokens on Polygon will be named UST with the different addresses. The price of USTw would be changed in the LP calc in quickswapLpPools.

shadoweddev:
okay, so the API doesn't pull token price from oracleID in quickLpPools. I initially thought oracleID mapped to an address in the book address which then API use to calculate price.
Rather , it actually pulls token price from the address given in lp0 and lp1 of a given entry in quickLpPools. 
Therefore I will adjust USTw to map to UST price in quickLpPools.

PR #586, opened 2022-01-20
Add Popsicle Finance to Polygon

AABeely:


packages/address-book/address-book/polygon/tokens/tokens.ts
```diff
@@ -821,6 +820,18 @@ const _tokens = {
       'Iron Finance is building a full suite of decentralized finance ecosystem of products and use cases on the Polygon network. The official Iron Finance token is the ICE token, inspired by A Song of Ice and Fire novels, where the Iron Throne is a well-known symbol.',
     logoURI: 'https://app.iron.finance/static/media/ice-square.6e960434.svg',
   },
+  PopsicleICE: {
```
roman-monk, line 823: if we call it like this zap into LP will not work. 
not sure how to call, i would rename Iron ICE to something like `ironICE`, or `ICEi`, and make this ICE to be default `ICE`, because its more popular. 
means need to rename old iron ICE in assets and images.

need to discuss it in discord with community.

AABeely: @roman-monk Thanks for looking over this. I've raised this in the strategists channel and been cleared to use ironICE naming for the Iron Finance token and pushed the changes up to both APP and API branches.

MirthFutures:
@AABeely Can you also mark the symbol name to match ironICE. Thats what the app reads.

AABeely:
@MirthFutures  I have updated the symbol name to ironICE in the address book.

PR #594, opened 2022-01-25
Add xBOO, xSCREAM, dQUICK

shatterproof:
0xDAO has unique types of pools where staked tokens are staked again in order to receive the rewards.

The prices of these tokens (`xBOO` and `xSCREAM`) were needed for the tokens oracle API, which this PR accomplishes by calculating their values based on the prices of their unstaked forms.

Additionally the APY calculation for the Spooky `BOO` single stake has been optimized to use the newly added `xBOO` price from the oracle now rather than needing to calculate it twice.


MirthFutures:
Theres no pricing for xBOO

src/utils/fetchXPrices.js
```diff
@@ -0,0 +1,31 @@
...
+  const totalXSupply = new BigNumber(await xTokenContract.methods.totalSupply().call());
+
+  return stakedInXPool.times(tokenPrice).dividedBy(totalXSupply).toNumber();
+};
+
+const fetchXPrices = async tokenPrices => {
+  return {
+    xBOO: await getXPrice(tokenPrices.BOO, BOO.address, xBOO.address),
+    xSCREAM: await getXPrice(tokenPrices.SCREAM, SCREAM.address, xSCREAM.address),
+    fBEETS: await getXPrice(tokenPrices.BEETS, fBEETSPool[0].address, fBEETS.address),
```
roman-monk, line 27: not bad, but few problems
- only for Fantom, so it can't be in general getAmmPrices
- no multicall, instead of 1 request (fetchMooPrices) there are 6 requests.
- all request are sequential, not parallel
- fBEETS should be based on BEETS-FTM LP price, not single BEETS?

shatterproof: Very helpful feedback, will have it fixed up tomorrow

shatterproof:
- Supports multiple chains to allow it to be used in general getAmmPrices
- Uses multicall for all calls on the same chain instead of multiple requests
- Requests are done in parallel now instead of sequential
- Removed fBEETS since it requires a varied formula (can revisit in the future)

roman-monk:
very nice

PR #607, opened 2022-02-01
PangolinV2 dual rewards apy calculation

pumpingGhost:
I implemented an apy calculation for dual reward farms on Pangolin.

The first pool to use this script will be LOOT-AVAX (PNG + LOOT rewards)
Want: 0x1bFae75b28925BF4a5bf830c141EA29CB0A868F1
PoolId: 65

Any further dual reward farms such as the upcoming UST pools can simply be added to the PangolinV2DualLpPools.json.

Implementation details:
The yearlyRewardsB are calculated by multiplying the yearlyRewardsA(PNG rewards, 18 dec) with the multiplier(divided by 1e18) provided by the rewarder contract.

### DIFF
--- a/src/api/stats/avax/getPangolinV2DualApys.js
+++ b/src/api/stats/avax/getPangolinV2DualApys.js
@@ -8,7 +8,7 @@ const ERC20 = require('../../../abis/ERC20.json');
 const fetchPrice = require('../../../utils/fetchPrice');
 const pools = require('../../../data/avax/pangolinV2DualLpPools.json');
 const { BASE_HPY, AVAX_CHAIN_ID } = require('../../../constants');
-const { getTradingFeeAprSushi } = require('../../../utils/getTradingFeeApr');
+const { getTradingFeeApr } = require('../../../utils/getTradingFeeApr');
 import { getFarmWithTradingFeesApy } from '../../../utils/getFarmWithTradingFeesApy';
 const { pangolinClient } = require('../../../apollo/client');
 const { compound } = require('../../../utils/compound');
@@ -18,7 +18,6 @@ const oracleIdA = 'PNG';
 const oracleA = 'tokens';
 const DECIMALSA = '1e18';
 
-const secondsPerBlock = 1;
 const secondsPerYear = 31536000;
 
 const liquidityProviderFee = 0.0025;
@@ -31,14 +30,10 @@ const getPangolinV2DualApys = async () => {
 
   const tokenPriceA = await fetchPrice({ oracle: oracleA, id: oracleIdA });
   const { rewardPerSecond, totalAllocPoint } = await getMasterChefData();
-  const { balances, allocPoints, rewarders } = await getPoolsData(pools);
+  const { balances, allocPoints, multipliers } = await getPoolsData(pools);
 
   const pairAddresses = pools.map(pool => pool.address);
-  const tradingAprs = await getTradingFeeAprSushi(
-    pangolinClient,
-    pairAddresses,
-    liquidityProviderFee
-  );
+  const tradingAprs = await getTradingFeeApr(pangolinClient, pairAddresses, liquidityProviderFee);
 
   for (let i = 0; i < pools.length; i++) {
     const pool = pools[i];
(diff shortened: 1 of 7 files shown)


src/api/stats/avax/getPangolinV2DualApys.js
```diff
@@ -0,0 +1,145 @@
...
+const getPangolinV2DualApys = async () => {
+  let apys = {};
+  let apyBreakdowns = {};
+
+  const tokenPriceA = await fetchPrice({ oracle: oracleA, id: oracleIdA });
+  const { rewardPerSecond, totalAllocPoint } = await getMasterChefData();
+  const { balances, allocPoints, rewarders } = await getPoolsData(pools);
+
+  const pairAddresses = pools.map(pool => pool.address);
+  const tradingAprs = await getTradingFeeAprSushi(
```
roman-monk, line 37: tradingAprs aren't calculated, but if you change it to just `getTradingFeeApr` it will be ok

src/api/stats/avax/getPangolinV2DualApys.js
```diff
@@ -0,0 +1,145 @@
...
+    liquidityProviderFee
+  );
+
+  for (let i = 0; i < pools.length; i++) {
+    const pool = pools[i];
+
+    const lpPrice = await fetchPrice({ oracle: 'lps', id: pool.name });
+    const totalStakedInUsd = balances[i].times(lpPrice).dividedBy('1e18');
+
+    const poolBlockRewards = rewardPerSecond.times(allocPoints[i]).dividedBy(totalAllocPoint);
```
roman-monk, line 49: i would just drop these `poolBlockRewards`, `secondsPerBlock` variables, as its not by blocks here

src/api/stats/avax/getPangolinV2DualApys.js
```diff
@@ -0,0 +1,145 @@
...
+    const yearlyRewards = poolBlockRewards.dividedBy(secondsPerBlock).times(secondsPerYear);
+    const yearlyRewardsAInUsd = yearlyRewards.times(tokenPriceA).dividedBy(DECIMALSA);
+    const yearlyRewardsBInUsd = await (async () => {
+      if (rewarders[i] === '0x0000000000000000000000000000000000000000') {
+        return 0;
+      } else {
+        const tokenPriceB = await fetchPrice({ oracle: pool.oracleB, id: pool.oracleIdB });
+        const rewarderContract = new web3.eth.Contract(Rewarder, rewarders[i]);
+        const tokenBMultiplier = new BigNumber(
+          await rewarderContract.methods.getRewardMultipliers().call()
```
roman-monk, line 59: this is not good place for another eth call, it should be included in getPoolsData multicall.
otherwise if it grows to 100 pools, we will have 100 calls here, instead of 1 with multicall.

i understand there is no rewarderAddress at the time of first call, so there are 2 ways:
1) if rewarder is immutable in MC, we can hardcode it in pools.json and get multiplier in multicall.
2) if its mutable, we can get rewarder in multicall, and then make another multicall inside getPoolsData to get list of multipliers, then we return { balances, allocPoints, rewardMultiplers } from getPoolsData, we dont really need to return rewarder from there. could return 0 if no rewarder

also there can be multiple rewards, not sure if it ever will

src/api/stats/avax/getPangolinV2DualApys.js
```diff
@@ -0,0 +1,156 @@
...
+
+  rewarders.forEach(rewarder => {
+    let rewarderContract = new web3.eth.Contract(Rewarder, rewarder);
+    let multiplier = rewarderContract.methods.getRewardMultipliers();
+    multipliersCalls.push({
+      multiplier: multiplier,
+    });
+  });
+
+  const multipliers = await multicall.all([multipliersCalls]);
```
roman-monk, line 151:
```suggestion
  const multipliers = (await multicall.all([multipliersCalls]))[0].map(m =>
    m.multiplier.map(v => new BigNumber(v))
  );
```

src/api/stats/avax/getPangolinV2DualApys.js
```diff
@@ -0,0 +1,156 @@
...
+    const totalStakedInUsd = balances[i].times(lpPrice).dividedBy('1e18');
+
+    const relativeRewards = rewardPerSecond.times(allocPoints[i]).dividedBy(totalAllocPoint);
+    const yearlyRewards = relativeRewards.times(secondsPerYear);
+    const yearlyRewardsAInUsd = yearlyRewards.times(tokenPriceA).dividedBy(DECIMALSA);
+    var yearlyRewardsBInUsd = new BigNumber(0);
+    var yearlyRewardsCInUsd = new BigNumber(0);
+
+    if (multipliers[i][0] != null) {
+      const tokenPriceB = await fetchPrice({ oracle: pool.oracleB, id: pool.oracleIdB });
```
roman-monk, line 51:
```suggestion
      let multiplier = multipliers[i][0];
      const tokenPriceB = await fetchPrice({ oracle: pool.oracleB, id: pool.oracleIdB });
      const yearlyRewardsB = yearlyRewards.times(multiplier).dividedBy(pool.decimalsB);
      yearlyRewardsBInUsd = yearlyRewardsB.times(tokenPriceB).dividedBy(pool.decimalsB);
```

src/api/stats/avax/getPangolinV2DualApys.js
```diff
@@ -0,0 +1,156 @@
...
+  for (let i = 0; i < pools.length; i++) {
+    const pool = pools[i];
+
+    const lpPrice = await fetchPrice({ oracle: 'lps', id: pool.name });
+    const totalStakedInUsd = balances[i].times(lpPrice).dividedBy('1e18');
+
+    const relativeRewards = rewardPerSecond.times(allocPoints[i]).dividedBy(totalAllocPoint);
+    const yearlyRewards = relativeRewards.times(secondsPerYear);
+    const yearlyRewardsAInUsd = yearlyRewards.times(tokenPriceA).dividedBy(DECIMALSA);
+    var yearlyRewardsBInUsd = new BigNumber(0);
```
roman-monk, line 47: var to let

src/api/stats/avax/getPangolinV2DualApys.js
```diff
@@ -0,0 +1,156 @@
...
+    var yearlyRewardsCInUsd = new BigNumber(0);
+
+    if (multipliers[i][0] != null) {
+      const tokenPriceB = await fetchPrice({ oracle: pool.oracleB, id: pool.oracleIdB });
+      const tokenBMultiplier = multipliers[i].map(v => v.multiplier[0]);
+      const yearlyRewardsB = yearlyRewards.times(tokenBMultiplier).dividedBy(pool.decimalsB);
+      yearlyRewardsBInUsd = yearlyRewardsB.times(tokenPriceB).dividedBy(pool.decimalsB);
+    }
+    if (multipliers[i][1] != null) {
+      const tokenPriceB = await fetchPrice({ oracle: pool.oracleB, id: pool.oracleIdC });
```
roman-monk, line 57: please rewrite this block more carefully, without B values

src/api/stats/avax/getPangolinV2DualApys.js
pumpingGhost: Thanks for the thorough feedback. I relocated the mapping as you suggested

src/api/stats/avax/getPangolinV2DualApys.js
pumpingGhost: Didn't know let was mutable.
In languages I am used to, let is always immutable, independent of the context, so I figured I'd have to use var.
Well, learned something new.

src/api/stats/avax/getPangolinV2DualApys.js
pumpingGhost: The implementation for the C token was actually work in progress, I should have marked the PR as draft.

I finished the implementation now and both additional rewards are properly calculated now.
For the 'oracle' variable I reused oracleB, as it's 'tokens' anyway, so I suggest we don't introduce a redundant property 'oracleC' to the pangolinV2DualLpPools.json.

pumpingGhost:
@roman-monk And thank you in general for the precise feedback. I'm new to node.js, but I'm giving it my very best to learn and get up to your standards.

roman-monk:
looks very nice now, limited to 2 rewards but its ok 😅

roman-monk:
if you have a courage maybe you could refactor `getJoeDualLpApys`? 🙂
i just found it also has `tokenPerSec` in for loop, that means we sequentially wait for every pool to get tokenPerSec, not even `Promises.all`

pumpingGhost:
> looks very nice now, limited to 2 rewards but its ok sweat_smile

I was already thinking of a way to handle this. Iterating over the multipliers wouldn't be a problem, but for the oracleIds we'd have to dynamically parse 'oracleId{X}'.
But I don't think we'll see any quadruple reward pools soon :smile: 

> if you have a courage maybe you could refactor `getJoeDualLpApys`? slightly_smiling_face i just found it also has `tokenPerSec` in for loop, that means we sequentially wait for every pool to get tokenPerSec, not even `Promises.all`

Yeah sure, can do that. Shall we stick to dual rewards for traderJoe or also support triples? As far as I know, there are no triple reward farms on joe.

roman-monk:
> I was already thinking of a way to handle this. Iterating over the multipliers wouldn't be a problem, but for the oracleIds we'd have to dynamically parse 'oracleId{X}'. But I don't think we'll see any quadruple reward pools soon 😄

i think oracleB (and C, ...) was a poor choice initially. 
would suggest to use array of rewards instead 
```
"rewards": [
  {
    "oracle": "tokens",
    "oracleId": "AVAX",
    "decimals": "1e18"
  },
    "oracle": "tokens",
    "oracleId": "UST",
    "decimals": "1e18"
  }
]
```
so we can iterate over rewarders and rewards.

like in curve pools tokens struct https://github.com/beefyfinance/beefy-api/blob/229d0b1d8bc760a5988341c45ed826927de3e05d/src/data/arbitrum/curvePools.json#L16

pumpingGhost:
> if you have a courage maybe you could refactor `getJoeDualLpApys`? slightly_smiling_face i just found it also has `tokenPerSec` in for loop, that means we sequentially wait for every pool to get tokenPerSec, not even `Promises.all`

I'd set up a new PR for this, if this is ok. Won't get to it in the next few hours.

PR #611, opened 2022-02-02
Beethoven demeters degree

AABeely:
Add support for the BeethovenX Demeter's Degree dual reward pool. Changes also include upgrading an existing vault that doesn't currently display the extra APY from dual rewards.


roman-monk:
looks good!
but what about our current `beets-two-became-one` vault? it doesnt compound HND currently

AABeely:
@roman-monk Yeah I noticed that the API for the vault wasn't returning APY figures including the HND rewards but by moving the pool into the new Dual rewards getter this should hopefully be resolved. As for whether it compounds those rewards I've not looked into it in depth but a quick look at a harvest transaction (https://ftmscan.com/tx/0xc8142d1aa2b04653b9938595304fd8002bf85f455161f769cfa10fdc87ef6ddf) seems to suggest the HND is being sold for FTM and compounded back into the pool correctly

roman-monk:
oh indeed, i checked wrong strat, so we under-reporting APY now?

src/api/stats/fantom/getBeethovenxDualApys.js
```diff
@@ -0,0 +1,110 @@
...
+      .dividedBy(DECIMALSA);
+    const yearlyRewardsBInUsd = await (async () => {
+      const masterchefContract = new web3.eth.Contract(MasterChefAbi, masterchef);
+
+      if (rewarders[i] === '0x0000000000000000000000000000000000000000') {
+        return 0;
+      } else {
+        const tokenPriceB = await fetchPrice({ oracle: pool.oracleB, id: pool.oracleIdB });
+        const rewarderContract = new web3.eth.Contract(BeethovenRewarder, rewarders[i]);
+        const tokenBPerSec = new BigNumber(await rewarderContract.methods.rewardPerSecond().call());
```
roman-monk, line 62: but this `rewardPerSecond` should be inside of multicall in getPoolsData

AABeely:
@roman-monk Yeah the `beets-two-became-one` vault is currently under-reporting APY

src/api/stats/fantom/getBeethovenxDualApys.js
roman-monk: see almost identical comment here https://github.com/beefyfinance/beefy-api/pull/607#discussion_r797643455

PR #622, opened 2022-02-07
Add tracking for Aurora BIFI/WETH

DefiDebauchery:


roman-monk:
we merge master into prod locally without PR to keep git history same

PR #632, opened 2022-02-12
TraderJOE HON + AVAX

pahor167:


roman-monk:
need to add HON token

pahor167:
> need to add HON token

HON token added

PR #633, opened 2022-02-12
Add missing BIFI Maxis

sircryptomaxi:


MirthFutures:
One other thing we need to do is put the Maxi address in governancePool.json file for each chain

sircryptomaxi:
> One other thing we need to do is put the Maxi address in governancePool.json file for each chain

Added in latest commit

PR #663, opened 2022-03-04
Implemented APY calculation for sJOE(successor of xJOE)

pumpingGhost:
sJOE is a revenue share model that replaced xJOE. There is no receipt token such as xJOE anymore, rather the JOE is staked in a contract that rewards in USDC once per day. The USDC rewards are airdropped to the staking contract once per day and the amount depends on the trading fee revenue of the platform. 

This made the APY calculation a bit more tricky than usual, because there is no definitive reward rate. Rather the staking contract keeps track of the totalAccumulatedRewards(all USDC rewards collected since the beginning of the farm). This property accRewardPerShare is written to once per day where the new rewards are added up.
```        
         accRewardPerShare[_token] = accRewardPerShare[_token].add(
            _accruedReward.mul(ACC_REWARD_PER_SHARE_PRECISION).div(_totalJoe)
        );
```

Since there is no property that gives a fixed reward rate, the APY needs to be calculated based on the past rewards. TraderJoe also states on their site that they give an APR calculation based on an average of the last 7 days. 

Since we don't have access to those daily reward numbers, my approach is to calculate an overall average per day by dividing the total number of accumulated rewards since beginning of the farm by the number of days that the farm exists.

What took me some time to figure out is how the ACC_REWARD_PER_SHARE_PRECISION works. It is used for scaling the rewardsPerShare based on the amount of JOE(18 dec) to the USDC value of that(6 dec). (See code snippet above)

I'm aware that the resulting APY is currently higher than what's displayed on TraderJoe, but we do not know how exactly they calculate their 7 days average and this calculation here is based on the true on-chain numbers.

### DIFF
--- a/packages/address-book/address-book/avax/tokens/tokens.ts
+++ b/packages/address-book/address-book/avax/tokens/tokens.ts
@@ -15,6 +15,30 @@ const AVAX = {
 } as const;
 
 const _tokens = {
+  ECD: {
+    name: 'Echidna ECD',
+    symbol: 'ECD',
+    address: '0xeb8343D5284CaEc921F035207ca94DB6BAaaCBcd',
+    chainId: 43114,
+    decimals: 18,
+    logoURI:
+      'https://raw.githubusercontent.com/traderjoe-xyz/joe-tokenlists/main/logos/0xeb8343d5284caec921f035207ca94db6baaacbcd/logo.png',
+    website: 'https://www.echidna.finance/',
+    description:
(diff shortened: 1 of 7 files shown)


roman-monk:
They get volume for last 7 days from subgraph `https://api.thegraph.com/subgraphs/name/traderjoe-xyz/exchange/graphql` and calc rewards as 0.05% of it.
This will match APR pretty closely.

pumpingGhost:
> 

Thanks for the tip Roman, didn't know you can query literally their whole database over the subgraph api :smile:

PR #687, opened 2022-03-16
small updates to contrib.md  

andreb0x:


andreb0x:
@sirbeefalot wanted to ensure you are okay with me contributing here since you set this doc up. i just went through the process of learning to modify API so just adding in some extra content that might help other beginners.

PR #704, opened 2022-03-26
WIP: Add quickswap tel eth vault

prevostc:


andreb0x:
NB vault APR and APY are NULL

prevostc:
> NB vault APR and APY are NULL

Fixed, definition was not in the right file

PR #705, opened 2022-03-27
xVOLT apy calculation

pumpingGhost:


src/api/stats/fuse/getVoltageApys.js
```diff
@@ -32,15 +45,21 @@ const getVoltageDualApys = async () => {
 
   const tokenPriceA = await fetchPrice({ oracle: oracleA, id: oracleIdA });
   const { rewardPerSecond, totalAllocPoint } = await getMasterChefData();
-  const { balances, allocPoints, tokenPerSecData } = await getPoolsData(pools);
 
   const pairAddresses = pools.map(pool => pool.address);
   const tradingAprs = await getTradingFeeApr(fusefiClient, pairAddresses, liquidityProviderFee);
 
+  pools.unshift(xVOLT);
```
roman-monk, line 52: something more explicit is better. but beside that, wont it add this object every time this function called?

pumpingGhost: Yeah you're right. I adapted it to create a new array containing the lpPools and xVOLT using the spread operator.

PR #706, opened 2022-03-27
Add DarkCrypto (NESS rewarder) apy calculation

pumpingGhost:
DarkCrypto has extended their portfolio of farms with four new farms that reward in NESS compared to the previous farms that reward in SKY.
https://www.darkcrypto.finance/bank


roman-monk:
this should be Launchpool

pumpingGhost:
Marked as draft until launchpool is ready.

PR #710, opened 2022-03-29
ncash avax 

andreb0x:


MirthFutures:
Can you work with Frondoto to get an accurate token description?

andreb0x:
changed to "NCash is the native cryptocurrency and protocol token of the Nitro Network. Nitro Network is building a world of private communication networks powered by IoT together with LoRaWAN/3G/4G and 5G."

PR #724, opened 2022-04-03
add stargate avax LP tokens

andreb0x:


MirthFutures:
Have to also add in fetchStargatePrices

PR #742, opened 2022-04-14
Web3 contract leak + axios removal + redis [REVIEW PLS]

seguido:
Fix issue caused by web3 contracts instantiated by web3.eth.contract, now uses raw contract object inside web3-eth-contract package.

### DIFF
--- a/src/api/stats/getApys.js
+++ b/src/api/stats/getApys.js
@@ -12,6 +12,7 @@ const { getAuroraApys } = require('./aurora');
 const { getFuseApys } = require('./fuse');
 const { getMetisApys } = require('./metis');
 const { getMoonbeamApys } = require('./moonbeam');
+const { getKey, setKey } = require('../../utils/redisHelper');
 
 const INIT_DELAY = process.env.INIT_DELAY || 60 * 1000;
 var REFRESH_INTERVAL = 15 * 60 * 1000;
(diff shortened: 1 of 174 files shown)


src/api/stats/getChainTvl.js
```diff
@@ -25,11 +26,12 @@ const getChainTvl = async chain => {
     const vaultBal = vaultBalances[i];
     let tokenPrice = 0;
     try {
+      // tokenPrice = 15.5;
       tokenPrice = await fetchPrice({ oracle: vault.oracle, id: vault.oracleId });
     } catch (e) {
       console.error('getTvl fetchPrice', chainId, vault.oracle, vault.oracleId, e);
     }
-    const tvl = vaultBal.times(tokenPrice).dividedBy(10 ** (vault.tokenDecimals ?? 18));
+    const tvl = vaultBal.times(tokenPrice).shiftedBy(vault.tokenDecimals ?? 18);
```
prevostc, line 34: not 100% certain but I think this should be shiftedby a negative number to get the same result

roman-monk: yeah should be negative 👍

src/api/stats/getApys.js
```diff
@@ -78,13 +81,31 @@ const updateApys = async () => {
     }
 
     console.log('> updated apys');
+    await saveToRedis();
   } catch (err) {
     console.error('> apy initialization failed', err);
   }
+  console.log(count);
+  if (--count == 0) {
```
roman-monk, line 89: what is this for?

PR #747, opened 2022-04-16
Clean up AVAX USDT tokens

cartman-dev:
In conjunction with beefyfinance/beefy-app#1177, these changes will clean up the various USDT tokens used on the AVAX network.
* `USDTe` remains unchanged
* Address book contained a 'USDT-Tether USD' duplicate entry that is removed
* `USDt`, the Native Tether USD on Avax, changed to `USDT`
* Legacy instances of `USDT` are changed to `USDTo` which was already in use on some legacy (EOL) vaults.  This is not active anywhere on the platform.
* Token descriptions and names have been updated to include whether they are bridged and from where
* Several pools that are no longer present (EOL or otherwise) have been removed

### DIFF
--- a/src/data/avax/comAvaxLpPools.json
+++ b/src/data/avax/comAvaxLpPools.json
@@ -71,7 +71,7 @@
     "lp1": {
       "address": "0xde3A24028580884448a5397872046a019649b084",
       "oracle": "tokens",
-      "oracleId": "USDT",
+      "oracleId": "USDTo",
       "decimals": "1e6"
     }
   },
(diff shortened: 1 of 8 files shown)


src/data/avax/comAvaxLpPools.json
```diff
@@ -71,7 +71,7 @@
     "lp1": {
       "address": "0xde3A24028580884448a5397872046a019649b084",
       "oracle": "tokens",
-      "oracleId": "USDT",
+      "oracleId": "USDTo",
```
roman-monk, line 74: this pool seems to be removed too

PR #763, opened 2022-04-28
Adapt SpookyV2LpApys + add new pools

pumpingGhost:
For the first two SpookySwapV2 farms, these two shared a ComplexRewarder for the second reward. This contract used allocPoints to split the rewards between the two farms, however it was lacking a totalAllocPoints interface. 
That is why a temporary solution had been implemented to iterate over all rewarders and sum up their allocPoints:
```
 // totalAllocPoint is not a public variable on the complex rewarder
 // find shared rewarders and sum their allocPoints
  ```

For the third V2 farm that is sFTMx-FTM, SpookySwap fixed their issue and deployed a new ComplexRewarder that contains the totalAllocPoints interface. 
In order to find a clean solution that works with the new ComplexRewarder and also supports the first two farms without the need for a distinct script, I optimized the script towards the new ComplexRewarder containing the totalAllocPoints interface and provided the first two pools with a static property for the totalAllocPonts of their rewarder, which is then picked up within the script.

Additionally, I made some adjustments to make SpookyV2LpApys work with traditional single reward farms as those will also be added to the masterchefV2 for future farms (see USDC-SD).

### DIFF
--- a/packages/address-book/address-book/fantom/tokens/tokens.ts
+++ b/packages/address-book/address-book/fantom/tokens/tokens.ts
@@ -38,6 +38,26 @@ const FTM = {
 } as const;
 
 const _tokens = {
+  sFTMx: {
+    name: 'sFTMx sFTMx',
+    symbol: 'sFTMx',
+    address: '0xd7028092c830b5C8FcE061Af2E593413EbbC1fc1',
+    chainId: 250,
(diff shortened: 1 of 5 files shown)


roman-monk:
seems working good 👍

PR #774, opened 2022-05-04
Fix beets "A diabolical balance"

pumpingGhost:
image
The farm for "A diabolical balance" has received an additional reward with "FHM". Therefore, I migrated it to a double rewards entry.

The boo-wftm-fhm pool is just an unincentivized LP, which is used for calculating the FHM price. If we prefer to fetch the price from coingecko in this case, please let me know.


src/data/fantom/spookyLpPools.json
```diff
@@ -1,4 +1,23 @@
 [
+  {
+    "name": "boo-wftm-fhm",
```
roman-monk, line 3: it is `dai`, also can update 1e09

pumpingGhost: My bad! It's fixed now.

src/data/fantom/spookyLpPools.json
```diff
@@ -1,4 +1,23 @@
...
+  {
+    "name": "boo-dai-fhm",
+    "address": "0xd77fc9c4074b56Ecf80009744391942FBFDDd88b",
+    "decimals": "1e18",
+    "poolId": 0,
+    "chainId": 250,
+    "lp0": {
+      "address": "0x8D11eC38a3EB5E956B052f67Da8Bdc9bef8Abf3E",
+      "oracle": "tokens",
+      "oracleId": "DAI",
```
MirthFutures, line 11: For dummy pools, move them here https://github.com/beefyfinance/beefy-api/blob/master/src/data/archive/oldLpPools.json

PR #777, opened 2022-05-05
Fix BSC BSHARE images

cartman-dev:
* Update address-book to use ID of BOMBSHARE for BSHARE on BSC, in order to differentiate from BSHARE (Based) on Fantom
* https://github.com/beefyfinance/beefy-app/issues/1219


cartman-dev:
v2 app code is going to change instead of this

PR #783, opened 2022-05-08
Add pFTM and pAVAX single pools

pumpingGhost:
Initially these single stake farms had very low TVL (<50k), but as the project has grown they reached 420k (Fantom) and 220k (Avax). Considering our partnership with Ripae, I think it would be beneficial to offer vaults for these farms.

The app PR is held back until the launchpool agreements are settled.


roman-monk:
still very low, 200k farm shouldn't be considered, EPETE is working weekly to clean such low tvl farms

PR #813, opened 2022-06-01
Add/babyswap pools

lodelux:
-add some babyswap pools
-add babyswap in bsc/platforms
-add babyswap in add-farm script


roman-monk:
very low tvl farms, will end up eol-ed by EPETE

PR #827, opened 2022-06-21
LP breakdown data

seguido:
Adds breakdown data for every lp we support on our /lps endpoint.


src/data/aurora/rosePools.json
```diff
@@ -2,28 +2,80 @@
...
+    "rewardPool": "0x226991aADeEfDe03bF557eF067da95fc613aBfFc",
+    "tokens": [
+      {
+        "oracle": "tokens",
+        "oracleId": "MAI",
+        "decimals": "1e18"
+      },
+      {
+        "oracle": "tokens",
+        "oracleId": "MAI",
```
roman-monk, line 16: this is basePool if not intentionally hardcoded to MAI

PR #829, opened 2022-06-23
Add MAI+3Pool polygon farm

BrknRobot:
Vault: 0x3CB928f9B49D9bD6eF43B7310dcC17dB0528CCc6
Strategy: 0x62068C5E65Df186BBEbC2B8d0D1a395eAB0ef9AE
Want: 0x447646e84498552e62eCF097Cc305eaBFFF09308
PoolId: 0

### DIFF
--- a/src/data/matic/curvePools.json
+++ b/src/data/matic/curvePools.json
@@ -159,7 +159,7 @@
     "tokens": [
       {
         "oracle": "tokens",
-        "oracleId": "mai",
+        "oracleId": "MAI",
         "decimals": "1e18"
       },
       {
--- a/src/data/matic/curvePools.json
+++ b/src/data/matic/curvePools.json
@@ -151,5 +151,10 @@
         "decimals": "1e18"
       }
     ]
+  },
+  {
+    "name": "crv-mai-3pool",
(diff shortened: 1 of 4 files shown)


src/data/matic/curvePools.json
```diff
@@ -151,5 +151,10 @@
         "decimals": "1e18"
       }
     ]
+  },
+  {
+    "name": "crv-mai-3pool",
```
roman-monk, line 156: lets rename to `curve-mai-3pool`, and please help to specify tokens array as we need it for incoming refactoring https://github.com/beefyfinance/beefy-api/blob/master/src/data/matic/curvePools.json#L141
seems it should be exactly same eurt, just different oracleId and decimals

src/data/matic/curvePools.json
```diff
@@ -151,5 +151,23 @@
...
     ]
+  },
+  {
+    "name": "curve-mai-3pool",
+    "pool": "0x447646e84498552e62eCF097Cc305eaBFFF09308",
+    "rewards": [],
+    "tokens": [
+      {
+        "oracle": "tokens",
+        "oracleId": "mai",
```
roman-monk, line 162: upper case please

PR #833, opened 2022-06-27
add BTC.b joe vault

lodelux:


packages/address-book/address-book/avax/tokens/tokens.ts
```diff
@@ -15,6 +15,17 @@ const AVAX = {
 } as const;
 
 const _tokens = {
+  'BTC.b': {
```
MirthFutures, line 18: adding the decimal might break the app, can you change the name to BTCb

lodelux: Fixed

PR #835, opened 2022-07-01
add vault biswap-sand-usdt

bowlingotter:


bowlingotter (commented):
adding SAND token to address book

PR #836, opened 2022-07-01
Handling for expired dual rewards in VVS

asdemt:
Calculations to not include expired dual reward in APY.

OLD INFO:

The pair seems to already exist on vvsDualLpPools.json. I didn't remove the entry there. Maybe the pool has rewarded dual rewards in the past, but currently it only rewards one token, VVS.

Vault: 0xA4D3044b9e790464A14c9801aBe39743d95498f2
Strategy: 0x9B43779d9223859821086430665c9C75Da039008
Want: 0x94FD83482016BCC5FcA5972A3635aA6524C3F557
PoolId: 20


roman-monk:
duplicated pool might work unpredicted, also there will be more pools like this.
the better fix is to handle `rewardEndTimestamp` in `getVvsDualApys` and return tokenPerSec as 0

asdemt:
> duplicated pool might work unpredicted, also there will be more pools like this. the better fix is to handle `rewardEndTimestamp` in `getVvsDualApys` and return tokenPerSec as 0

This PR no longer adds the new pool, but uses the existing one. Fixed APY calculation as suggested.

roman-monk:
awesome

PR #868, opened 2022-07-29
Add bolide finance

CryptoDev0203:


CryptoDev0203:
@roman-monk Can I use our BLID token from "blockchain-addressbook" API?

PR #876, opened 2022-08-04
[WIP][Request for Review] Add TS types to enable noImplicitAny

prevostc:
Follow up on pr [!873](https://github.com/beefyfinance/beefy-api/pull/873)

I enabled `noImplicitAny` and added types information when the TS checker complained.
This is a partial implementation as there is a lot of types to add and I feel like multiple small PR will be better than a huge one.

This work is preliminary to further refactoring so I tried to leave current logic as close to the original implementation as possible.


roman-monk:
lets not stress us that much `noImplicitAny` please. 
i even dont get why its needed to migrate all files to ts, but some guys like to write ts so it seems fine. 
however i think we dont need to strict it that much. 
it does slow me down to write and to read code, something like this i think is quite awful
```
function sortByKeys<TKey extends string, TValue, TType extends Record<TKey, TValue>>(
  o: TType
): TType {
```
majority of contributors in beefy-api would just write js, only chebin would push ts but hopefully without covering everything with types. i also found that js is much simpler for new strategists, so its just naturally to keep it in api.

PR #879, opened 2022-08-05
Add Aave v3 Optimism sUSD and AAVE

asteroidegen:


MirthFutures:
sUSD has a Supply cap

roman-monk:
aave would be nice vault

PR #882, opened 2022-08-06
Add Solar wstKSM-xcKSM

sircryptomaxi:


MirthFutures:
Needs to account for new fee.

PR #883, opened 2022-08-08
Add xcaUSD token to AB and Beamswap LPs

schnarkus:


MirthFutures:
Need to include "beefyFee": 0.095; in the lp data. Look at recent commits for guidance.

PR #884, opened 2022-08-09
[UPDATE] wbtce and add renBTC

erenjaegar77:


packages/address-book/address-book/avax/tokens/tokens.ts
```diff
@@ -3913,15 +3913,26 @@ const _tokens = {
     chainId: 43114,
     address: '0x50b7545627a5162F82A992c33b87aDc75187B218',
     decimals: 8,
-    name: 'Wrapped BTC',
-    symbol: 'WBTC',
+    name: 'Wrapped BTC (eth bridged)',
```
roman-monk, line 3916: why do we add this `(eth bridged)`? its not in token contract name.
same in USDT.e, but not in other .e tokens.
also all other .e tokens have the same id and symbol (symbol without `.`), though it seems incorrect but we had some issues when they dont match iirc

erenjaegar77: i push a new fix including usdt name too

PR #890, opened 2022-08-16
Set "boosted: true" for Solidly vaults

schnarkus:


roman-monk:
not good, dont copy-paste, just change `if (params.boosted)` to `if (params.boosted && params.NFTid)` in 2 places in original getSolidlyGaugeApys

roman-monk:
i see, dystopia rewardRate is 1e36, weird

PR #897, opened 2022-08-25
Add bolide-cakev2-blid-usdt

bolidetech:


roman-monk (changes requested):
should use common/getMasterChefApys and 0.095 fees. 
wont be merged anyway without safu report

PR #898, opened 2022-08-25
Add two Cone pools

schnarkus:


MirthFutures:
Added for you to avoid conflicts

PR #907, opened 2022-09-03
Add some dystopia and spiritV2 pools

schnarkus:


roman-monk:
i would strongly reconsider no-tvl, no-apr farms, are they even attractive to vault.
maybe consult with EPETE.
to not add more dead vaults 
image

PR #914, opened 2022-09-11
Add POOP-GLMR

pumpingGhost:


pumpingGhost:
> @pumpingGhost we need to be better with token descriptions. Can you put a better description of what the token is besides its a governance token? Maybe some info about the platform?

Yeah I agree, especially when the token is called POOP :smile: 
I added another description on the platform itself.

PR #915, opened 2022-09-12
Token endpoint

seguido:
Expose addressbook tokens from a new endpoint

`/tokens` returns all tokens by chain
`/tokens/:chain` returns all tokens for the given chain (example: `/tokens/bsc`


DefiDebauchery:
Let it be known that this is super-epic. Thanks!

PR #917, opened 2022-09-12
Add two Velodrome pools

cartman-dev:


packages/address-book/address-book/optimism/tokens/tokens.ts
```diff
@@ -398,6 +398,16 @@ const _tokens = {
...
     logoURI: '',
   },
+  USDP: {
+    name: 'USD+ Stablecoin',
+    symbol: 'USDP',
+    address: '0x73cb180bf0521828d8849bc8CF2B920918e23032',
+    chainId: 10,
+    decimals: 6,
+    website: 'https://overnight.fi/',
+    description: 'USD+ is USDC that pays you yield daily via rebase.',
```
MirthFutures, line 408: Same with this one. USD+ actually uses yield strategies to earn the yield. It would be helpful to talk more about what it is in the description.

cartman-dev:
```suggestion
    description: 'USD+ is USDC that pays you yield daily via rebase.  It is 100% collateralized with assets immediately convertible into USDC.  Yield is generated via strategies such as lending and stable-to-stable pools. Initial strategies include Aave, Rubicon, and Pika.',
```

PR #918, opened 2022-09-12
Update fees for APYs

seguido:
Remove all hardcoded performance fees being used and replace them from the values being retrieved from the vault contracts. Default for missing fees is 9.5%. 

@roman-monk We had a bunch of 4.5% fee values hardcoded across the API for pools that do not match a vault, therefore their performance fees are missing and would be assigned a 9.5% fee now. I'm not aware of any place where we need an internal APY to find another one, do you think this could be an issue? I think this are mostly old vaults or pools/lps that are used for price calculations but are also thrown into the apy finding codebase.

Was also thinking about removing some old platforms that shouldn't be used anymore, will keep expanding on this but would be great if you could take an initial look at this


roman-monk:
yeah we can remove old APYs https://github.com/beefyfinance/beefy-api/commit/9849f113021e4b234a0a4847edae715801c7ed30
less requests to RPC as well. 
we keep lpPools.jsons so users see correct deposit prices but likely we can remove them too

PR #927, opened 2022-09-20
Update Jarvis/Fiat price fetch

sircryptomaxi:
Added logic for fetching fiat currencies from CoinGecko as no reliable oracle for CAD could be sourced simply. CoinGecko has CADC as 70% over the actual value intermittenly.


sircryptomaxi:
Just now found the chainlink oracle for Forex, that would be more appropriate

PR #937, opened 2022-09-27
Add new spooky Chef to AB and create json for this new Chef farms

DerLomo:


roman-monk:
why `deus-` prefix though?

DerLomo:
> why `deus-` prefix though?

I was unsure of what convention is followed, but I assumed that as in the other json the first part is the reward token

PR #943, opened 2022-10-03
DEI-USDC, Update DEI and DEI_OLD

schnarkus:
It works now with Lomo's changes to the spooky V3 pairs. 

DEI


packages/address-book/address-book/fantom/tokens/tokens.ts
```diff
@@ -344,12 +344,12 @@ const _tokens = {
   DEI: {
     name: 'DEI',
     symbol: 'DEI',
-    address: '0xDE12c7959E1a72bbe8a5f7A1dc8f8EeF9Ab011B3',
```
roman-monk, line 347: we should keep old token as DEI_old or something, we have vaults with it

PR #945, opened 2022-10-05
Add Velo USDC-MIM, WETH-BIFI

schnarkus:


MirthFutures:
Thanks for adding the docs links @schnarkus

PR #952, opened 2022-10-18
Add WBTCwh-GLMR

pumpingGhost:


MirthFutures (commented):
@pumpingGhost We need to add a documentation key in the addressbook additions

PR #968, opened 2022-10-31
Add Sonne Platform and initial pools

cartman-dev:
Add Sonne platform to address-book and API.  Create common/getCompoundV2Apys.ts that can be used for multiple platforms.

### DIFF
--- a/src/api/stats/common/getCompoundV2Apys.ts
+++ b/src/api/stats/common/getCompoundV2Apys.ts
@@ -1,0 +1,214 @@
+import BigNumber, { BigNumber as BigNumberStatic } from 'bignumber.js';
+import Web3 from 'web3';
+import { AbiItem } from 'web3-utils';
+import { ChainId } from '../../../../packages/address-book/address-book';
+import { getTotalPerformanceFeeForVault } from '../../vaults/getVaultFees';
+import getBlockTime from '../../../utils/getBlockTime';
+import fetchPrice from '../../../utils/fetchPrice';
+import { compound } from '../../../utils/compound';
+import { getContractWithProvider } from '../../../utils/contractHelper';
+import { BASE_HPY } from '../../../constants';
+
+const IToken = require('../../../abis/VToken.json');
+const IComptroller: AbiItem[] = require('../../../abis/IComptroller.json');
+
+const SECONDS_PER_YEAR = 31536000;
+
+const getCompoundV2ApyData = async (params: CompoundV2ApysParams) => {
+  params.compOracle = params.compOracle || 'tokens';
+  params.compDecimals = params.compDecimals || '1e18';
+  params.comptrollerAbi = params.comptrollerAbi || IComptroller;
+  params.cTokenAbi = params.cTokenAbi || IToken;
+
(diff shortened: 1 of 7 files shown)


src/api/stats/common/getCompoundV2Apys.ts
```diff
@@ -0,0 +1,214 @@
...
+  params.comptrollerAbi = params.comptrollerAbi || IComptroller;
+  params.cTokenAbi = params.cTokenAbi || IToken;
+
+  const [calculatedBlockTime] = await Promise.all([getBlockTime(params.chainId)]);
+  params.secondsPerBlock = params.secondsPerBlock || calculatedBlockTime;
+
+  let apys = {};
+
+  let promises = [];
+  params.pools.forEach(pool => promises.push(getPoolApy(params, pool)));
```
roman-monk, line 29: would be super cool if its rewritten in a multicall way (check `common/getMasterChefApy`).
could be supplyCalls and borrowCalls with 1 multicall total vs current 40+

cartman-dev: I started there, didn't get it working, then came at it from a different direction. Now that the calls are right, I can bring it back to multicall.

cartman-dev:
@roman-monk As requested, updated to use multicall.

roman-monk:
amazing

src/api/stats/common/getCompoundV2Apys.ts
```diff
@@ -0,0 +1,271 @@
...
+
+  let promises = [];
+  params.pools.forEach(pool =>
+    promises.push(fetchPrice({ oracle: pool.oracle, id: pool.oracleId }))
+  );
+  const tokenPrices: BigNumber[] = await Promise.all(promises);
+
+  params.pools.forEach(pool => {
+    const cTokenContract = getContract(cTokenAbi, pool.cToken);
+    supplyRateCalls.push({ supplyRate: cTokenContract.methods.supplyRatePerBlock() });
```
roman-monk, line 142: https://github.com/beefyfinance/beefy-api/blob/704761bc1ac86357dd5ae7dfc54c1d1406b7ab69/src/api/stats/bsc/getDotDotApy.js#L33-L49

can call multiple methods in 1 call object. doesnt change anything and no need to update but could be handy next time. i think it looks simpler, thats why i suggested supplyCalls and borrowCalls

roman-monk:
🚨
response object is broken tho, how did you test it? 🙈
image

cartman-dev:
> 🚨
> 
> response object is broken tho, how did you test it? 🙈
> 
> image
> 
> 

Trusted the debug text obviously (oops).

src/api/stats/common/getCompoundV2Apys.ts
cartman-dev: Yes, I think that would probably have been cleaner.  Honestly, I just went with exactly the form that was done in `getMasterChefApys.ts` Another issue I ran into for Sonne specifically is that the `compSpeeds()` function we have used in the past (Scream, etc) returns zero for Sonne.  Maybe the next platform will be different again... Perhaps the dream of a common CompoundV2 implementation is too ambitious.

PR #971, opened 2022-11-01
Normalize all LP oracleIds to use native tokens + add missing jarvis token prices

seguido:
- Add missing jarvis tokens to support breakdown data on all LPs
- Remove wrapped versions of tokens from all LPs' oracleIds in favor of their native version (WBNB -> BNB)
- Set wrapped token with exact same value as their unwrapped version for consistent values across all pools right after pool price fetching


packages/address-book/address-book/polygon/tokens/tokens.ts
```diff
@@ -290,7 +290,7 @@ const _tokens = {
   PAR: {
     name: 'PAR Stablecoin',
     symbol: 'PAR',
-    address: '0x7b367a058f370c0057762280547d0b974cf3ac10',
+    address: '0xE2Aa7db6dA1dAE97C5f5C6914d285fBfCC32A128',
```
wivern-co-uk, line 293: Is it intentional change?

src/api/stats/getAmmPrices.ts
```diff
@@ -580,7 +585,35 @@ const updateAmmPrices = async () => {
       };
     };
 
-    const ammPrices = fetchAmmPrices(pools, knownPrices);
+    // All LP oracleIds should be set to the native assets, the wrapped version of each of those is then set based on those to avoid different values
+    const ammPrices = fetchAmmPrices(pools, knownPrices).then(prices => {
+      const gasTokens = [
```
wivern-co-uk, line 590: Suggestion to import data from address book, no need to list native coins manually

src/data/alpacaLpPools.json
```diff
@@ -14,7 +14,7 @@
     "lp1": {
       "address": "0xbb4CdB9CBd36B01bD1cBaEBF2De08d9173bc095c",
       "oracle": "tokens",
-      "oracleId": "WBNB",
+      "oracleId": "BNB",
```
wivern-co-uk, line 17: 1. These json files are semi-generated. 
2. Change like this adding manual work on plate of every vault developer which is hard to communicate and ensure. So data here inevitably will become inconsistent.

wivern-co-uk (changes requested):
Could we please unify price without patching json files?!

packages/address-book/address-book/polygon/tokens/tokens.ts
seguido: Yes, the wrong address was being used

src/api/stats/getAmmPrices.ts
seguido: There's no reference in the addressbook to a token being native

seguido:
> Could we please unify price without patching json files?!

@wivern-co-uk Price fetching from LP jsons sets prices based on the oracleId being used. We are not consistent with this, so some LPs are setting a price for BNB whilst others are setting the price for WBNB, according to the amounts of liquidity involved in the pools used for each different oracle we can end up with a wide spread between the result from both tokens. (Last week we were experiencing an 8% difference between AVAX and WAVAX).

These oracleId keys is how prices get resolved from the jsons, so unless they are the same we could end up with different prices for the "same" token being used on different places

src/api/stats/getAmmPrices.ts
seguido: we usually also have the native key referencing the wrapped token

`const BNB = {
  name: 'WBNB Token',
  symbol: 'WBNB',
  address: '0xbb4CdB9CBd36B01bD1cBaEBF2De08d9173bc095c',
  chainId: 56,
  decimals: 18,
  website: 'https://www.binance.com/',
  description:
    'Binance Coin (BNB) is an exchange-based token created and issued by the cryptocurrency exchange Binance. Initially created on the Ethereum blockchain as an ERC-20 token in July 2017, BNB was migrated over to Binance Chain in February 2019 and became the native coin of the Binance Chain.',
  logoURI:
    'https://tokens.pancakeswap.finance/images/0xbb4cdb9cbd36b01bd1cbaebf2de08d9173bc095c.png',
} as const;`

`BNB,
  WBNB: BNB,`

src/data/alpacaLpPools.json
seguido: The alternative I see is changing the keys once loaded

src/api/stats/getAmmPrices.ts
wivern-co-uk: Not entirely true. It's possible to get those via `WNATIVE`. Not perfect but better than having a hardcode here.

It's so much work to bring new networks, would be great to have one thing less to keep in mind.

src/data/alpacaLpPools.json
seguido: But it requires a similar type of tuning, getting all the native/wrapped tokens and setting oracleIds to the same. It would mean the jsons themselves aren't useful as they are and data is dropped

wivern-co-uk:
Completely agree price should be unified, however still see no reason to corrupt json files, which quite accurately presenting what's inside any particular LP.

Price for token is being chosen from LP with highest liquidity. 

Easy solution: If we can admit that major liquidity available for market in wrapped tokens then price value should be copied to native.

Bit harder: when comparing LPs liquidity consider native pairs too

seguido:
Sadly we can't make those kinds of assumptions, depending on the chain executing first, the order in which token0 and token1 are set, the oracleId(which can be added via scripts but we've found strategists also just copy other jsons manually) it's really hard to be right when assuming always one of both oracles would be correct.

I'm honestly leaning more towards keeping jsons as they are, since you're right, it will be really hard to prevent this from becoming inconsistent again and changing the lp-price-resolving bit of code to change all oracleIds to either wrapped OR unwrapped (we just need to pick one), and set the other once finished.

Does this sound good?

wivern-co-uk:
> lp-price-resolving bit of code

Sounds great to me! Will happily assist with code review.

seguido:
@wivern-co-uk wdyt?

src/api/stats/getAmmPrices.ts
```diff
@@ -580,7 +585,19 @@ const updateAmmPrices = async () => {
...
+    const ammPrices = fetchAmmPrices(pools, knownPrices).then(prices => {
+      //Set prices for the wrapped version of native tokens (if native was set)
+      const nativeTokens = new Set(
+        Object.values(addressBookByChainId).map(addressbook =>
+          addressbook.tokens.WNATIVE.symbol.slice(1)
+        )
+      );
+      nativeTokens.forEach(nativeToken => {
+        if (prices.tokenPrices.hasOwnProperty(nativeToken))
+          prices.tokenPrices[`W${nativeToken}`] = prices.tokenPrices[nativeToken];
```
wivern-co-uk, line 597: I'm quite confident that prices should be extracted from wrapped to native, not vice versa. My reasoning is based on opinion that most AMMs are Uniswap v2 clones which operates wrapped tokens. On another hand the only AMM working with native directly 1inch which is not present on all networks. 
Could you please check and post here pairs and their liquidities for data-driven decision. Should be double by console.log somewhere inside `fetchAmmPrices`. Agree?

seguido: I mean it's symbolic wether we use one or the other, it won't make a difference. Wether we initially set them all as native and later derive the wrapped value or vice versa is the same

seguido:
@roman-monk wdyt?

roman-monk (approved):
yeah i like original json patching instead of current implicit slicing, but still fine 👌

PR #972, opened 2022-11-03
Fix CompoundV2Apys response object

cartman-dev:
@roman-monk🙈


roman-monk:
still bad

image

it helps to launch it locally to test

cartman-dev:
Ok, I did simulate this before, but it took a few tries before my smooth brain latched on to the problem.  Should be good now.

PR #975, opened 2022-11-07
Add USD+ distinction between different chains

schnarkus:


MirthFutures:
@schnarkus I dont think you need to change the addressbook, just the keys in the price oracle

PR #983, opened 2022-11-26
Add Beam wstDOT-xcDOT

schnarkus:


roman-monk:
no APY, no lps prices work

PR #985, opened 2022-11-28
Add Beamswap multiRewardPools

schnarkus:


roman-monk:
`unsolved:  wstDOT xcDOT beamswap-wstdot-xcdot`
lets fix this

PR #987, opened 2022-12-04
swapfish dai usdc vault

seguido:
My first :hype_emoji:


packages/address-book/address-book/arbitrum/platforms/swapfish.ts
```diff
@@ -0,0 +1,4 @@
+export const swapfish = {
+  minichef: '0x33141e87ad2DFae5FBd12Ed6e61Fa2374aAeD029',
```
roman-monk, line 2: will be deprecated data in few months

PR #990, opened 2022-12-06
add optional oracleId to tokens in addressbook, and add it for FISH

seguido:
This would allow tokens with the same name to be displayed their names in the app whilst using a different price oracleId.

Current situation:
FISH (polycat - polygon) and FISH (swapfish - arbitrum) have identical names and the app uses their names as the oracleId for them.

However the oracleId for fish (swapfish) is SWAPFISH, being able to set this in the ab would allow the app to update the oracleId accordingly whilst keeping FISH as name.


seguido:
Alternative is to swap FISH entry in arbitrum to have key as SWAPFISH and symbol as FISH. This is hacky on as how the app is implemented and load addressbook data and can be a bit more obscure which is why I'd rather stick to the above method

PR #1045, opened 2023-02-25
Zap support endpoint (+other bits)

ReflectiveChimp:
Requires https://github.com/beefyfinance/beefy-v2/pull/1285

Add vault zap availability endpoint
Dev: Add more types
Dev: Abstract redis cache backend and add file based cache for local testing
Dev: Add event bus for communication between services
Dev: Add multicall v3 addresses per chain
Dev: Add chain/number/promise/big-number/error helpers


seguido (approved):

src/api/tokens/tokens.ts
```diff
@@ -0,0 +1,280 @@
...
+        id: vault.earnedToken,
+        symbol: vault.earnedToken,
+        name: vault.earnedToken,
+        chainId,
+        oracle: 'tokens', // ???
+        oracleId: vault.earnedOracleId || vault.earnedToken,
+        address: vault.earnedTokenAddress,
+        decimals: vault.earnedTokenDecimals || 18,
+      });
+    }
```
seguido, line 120: Is this reacheable? I don't believe we have any of these

seguido: just checked and we don't have a mention of 'native' in vault jsons

src/api/treasury/getTreasury.ts
```diff
@@ -17,128 +14,115 @@ import {
...
-            name: vault.token,
-            address: vault.tokenAddress,
-          };
-        }
-      }
+}
 
-      const vaultAddress = vault.earnContractAddress.toLowerCase();
-      chainTokens[vaultAddress] = {
+function getVaultAddressesByChain(): TreasuryAssetRegistry {
```
seguido, line 121: I think I'd rather have all mootokens loaded in the token service and maybe stored separately/flagged so they can be retrieved independently and "token loading" is fixed in that section of the api. wdyt?

src/api/tokens/tokens.ts
ReflectiveChimp: This would pick up gov/earnings vaults (if vault service didn't pre-filter them).

src/api/treasury/getTreasury.ts
ReflectiveChimp: The vaults endpoint adds PPFS which is used by treasury

src/api/tokens/tokens.ts
seguido: even if we had gov vaults there, we don't reference the earnedTokenAddress as native

seguido (approved):

PR #1048, opened 2023-03-02
[New platform] Equilibre on Kava

seguido:
@MirthFutures wanted to do a final revision on safuness I think so waiting on that or anyone willing to verify.


seguido:
Actually due to some errors emissions are kicked a whole week, no urgency

PR #1053, opened 2023-03-03
Use seconds per year when annualizing balRewardRate during aura-bal apy calc

### LINKED ISSUE #1052: Use seconds per year when annualizing balRewardRate during aura-bal apy calc
Aura say this
https://docs.aura.finance/developers/how-to-___/calculate-aprs-from-aura-pool-yield
image

But beefy say this
https://github.com/beefyfinance/beefy-api/blob/bf2c1a2deb6f7299f18c8524af4d2d45eda7f7e3/src/api/stats/ethereum/getAuraBalApy.js#L60

murderteeth:
https://docs.aura.finance/developers/how-to-___/calculate-aprs-from-aura-pool-yield

close #1052

### DIFF
--- a/src/api/stats/ethereum/getAuraBalApy.js
+++ b/src/api/stats/ethereum/getAuraBalApy.js
@@ -57,7 +57,7 @@ const getYearlyRewardsInUsd = async () => {
   const bbaUSDVirtualGauge = getContractWithProvider(IAuraGauge, bbaUSDVirtualGaugeAddress, web3);
   const bbaUSDRewardRate = new BigNumber(await bbaUSDVirtualGauge.methods.rewardRate().call());
   const auraData = await getAuraData();
-  const yearlyRewards = balRewardRate.times(3).times(BLOCKS_PER_DAY).times(365);
+  const yearlyRewards = balRewardRate.times(86_400 * 365);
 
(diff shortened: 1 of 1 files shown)


MirthFutures:
The calc is the same, the blocks in a day variable times 3 is equal to 86400.

murderteeth:
> The calc is the same, the blocks in a day variable times 3 is equal to 86400.

thanks @MirthFutures ..I totally missed that `BLOCKS_PER_DAY` was hardcoded at the top of the code. Should have been taking my time =)  

Now I see that most of the apy code has this pattern of setting a `BLOCKS_PER_DAY` const at the top of the page to 28800 then computing yearly reward with `rewardRate.times(3).times(BLOCKS_PER_DAY).times(365)`.

Why is this?  Using `SECONDS_PER_YEAR = 31_536_000` and `rewardRate.times(SECONDS_PER_YEAR)` is more intuitive I think.

cartman-dev:
@murderteeth one thing that comes to mind is that the contracts actually reward "per block" rather than "per second". The expectation is that block tempo is stable for the chain, but if it were to change all those "per second" values would be wrong. In this context, "per block" is more consistent. (As paired with a blocks per second conversion, which may or may not ever change)

PR #1090, opened 2023-04-01
Arbidex

cartman-dev:
APY is comparable to `getMasterChefApys.ts` with a couple exceptions:
* Rewards are fixed in ARX and WETH
* Some MasterChef calls are different due to the Arbidex contract
* No subgraph was found, so no trading fees (hopefully will change soon)

### DIFF
--- a/src/api/stats/arbitrum/getArbidexApys.ts
+++ b/src/api/stats/arbitrum/getArbidexApys.ts
@@ -1,0 +1,145 @@
+import BigNumber from 'bignumber.js';
+import { getContract, getContractWithProvider } from '../../../utils/contractHelper';
+import getApyBreakdown, { ApyBreakdownResult } from '../common/getApyBreakdown';
+import { getTradingFeeApr } from '../../../utils/getTradingFeeApr';
+import fetchPrice from '../../../utils/fetchPrice';
+import { MultiCall } from 'eth-multicall';
+import { multicallAddress } from '../../../utils/web3';
+import { arbitrumWeb3 as web3 } from '../../../utils/web3';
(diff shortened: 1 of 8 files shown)


src/api/stats/arbitrum/getArbidexApys.ts
```diff
@@ -0,0 +1,145 @@
...
+    }
+  }
+
+  return apys;
+};
+
+const getMasterChefData = async () => {
+  const masterchefContract = getContractWithProvider(abi, masterchef, web3);
+
+  const arxRewardsPerSec = new BigNumber(await masterchefContract.methods.arxPerSec().call());
```
roman-monk, line 111: this could be 1 multicall instead of 4 calls.
or ideally just combine it with existing multicall in getPoolsData
```js
const chefCalls = [{
  arxPerSec: masterchefContract.methods.arxPerSec(),
  WETHPerSec: masterchefContract.methods.WETHPerSec(),
  arxTotalAllocPoint: masterchefContract.methods.arxTotalAllocPoint(),
  WETHTotalAllocPoint: masterchefContract.methods.WETHTotalAllocPoint(),
}];

const res = await multicall.all([balanceCalls, allocPointCalls, chefCalls]);
const chefData = res[2].map(v => ({
  arxPerSec: new BigNumber(v.arxPerSec),
  // other fields
}))[0];
```

roman-monk (changes requested):

src/api/stats/arbitrum/getArbidexApys.ts
cartman-dev: @roman-monk I combined the chef calls and pool calls into a single multicall as you suggested.

PR #1145, opened 2023-06-13
[FEAT] RPC Multicall Abstraction

seguido:
Migrate RPC libraries to use viem. 
- enables us to abstract away from multicall code and get even better optimizations than we were seeing before.
- app-wide multicall batching, cross file requests can be multicalled, saving on rpc requests and costs
- type inference for rpc calls (yey!!!)
+ more cool stuff!

Complete migration of all price and vault related code to the new libraries.
Bump node version requirement to 18 to take full advantage of viem's features!

The way to work moving forward will be to call `fetchContract` and do all related calls via that method. If we were to run into a call that doesn't allow calls from other contracts (tx.origin != tx.sender), fetchNoMulticallContract can be used instead [try to avoid at all costs]

To take full advantage of type inference, ABIs must be migrated to ts files such as it's been done with all price related ones. Please take this into consideration when adding new ABI files, I'll be migrating all existing ones as I cover other sections of the API in future PRs.

In order to fully maximize multicall optimizations there's one single requirement: Please make all needed RPC calls before awaiting. For example
```
    const price = await getLpPrice(pool);
    const balances = await getLpTokenBalances(pool); // This call can't be added to the multicall till price one has been answered
```

```
    const [price, results] = await Promise.all([getLpPrice(pool), getLpTokenBalances(pool)]);
    // here both promises are awaited after being called, so they can be batched together
```


roman-monk:
```
src/api/stats/common/hop/getHopCommonApys.js:36:49
You must provide the json interface of the contract when instantiating a contract object.
```

src/utils/fetchAmmPrices.js
```diff
@@ -159,25 +161,28 @@ const fetchAmmPrices = async (pools, knownPrices) => {
 };
 
 const fetchChainPools = async (chain, pools) => {
-  const provider = new ethers.providers.JsonRpcProvider(MULTICHAIN_RPC[chain]);
-  const multicall = new ethers.Contract(MULTICALLS[chain], MulticallAbi, provider);
+  const multicallContract = fetchContract(MULTICALLS[chain], BeefyPriceMulticall, chain);
+  const lpInfos = await Promise.all(
```
roman-monk, line 165: will crash whole api here without try/catch if request failed?
need to catch and also fill .totalSupply, .lp0.balance, .lp1.balance with something so it wont crash later in calc.

same for fetchDmmPrices i guess

prevostc: probably better to catch and handle errors on the caller side instead of each and every method calling the rpc

seguido:
> surprisingly no big improvements for me, still takes quite some time and there are failed requests. some errors are very verbose, whole ABI is printed.
> 
> but overall its ok to merge and test in prod as long as there are no crashes. slightly unpleasant that api uses node v18 vs v16 in app, maybe can upgrade app as well.

Hey @roman-monk ! Thanks for taking a look at everything. I appreciate the feedback. 

The PR exclusively focuses on migrating to this new capabilities but I didn't optimize multicall batching settings to an optimal value, the idea was to get everyone to start writing code with this new methods so I don't have to go around migrating even more than necessary.

As it stands right now, client settings are:
```
batchSize: 128, // bytes
wait: 1500 //ms
```

Locally, running with this params I get a full price update in ~25 seconds. However the default batchSize is 1024 bytes and when running it with that change a full update is finished in ~6.5 seconds.

It's a matter of fully optimizing the settings, I'd rather run through that process once the entire api is migrated so we can adjust for our entire volume of calls. However, 128 bytes was a value I was using for testing and I will bump it to the default 1024 since it appears our rpcs can handle that payload without issues.

src/utils/fetchAmmPrices.js
seguido: these functions were already crashing when errors happened during rpc calls. They just crashed later when the fields that weren't loaded tried to be accessed... 

I'll look into rewriting everything somehow

seguido:
@roman-monk since fetchAmmPrices and fetchDmmPrices need every inner request to be successful in order to find prices correctly, I add a promise retrying function that retries the call we provide N times with an exponentially increasing interval in between attempts. 

Awesome feature coming from this is that promises being retried can get multicalled with others due to our global multicall aggregation.

I'm stil debating internally on whether we want every rpc call in the api to benefit from this feature or not, wdyt?

ps: I tried to cut back on error logs sizes a bit

roman-monk:
fetchAmm was pretty stable i think, it wouldn't crash if some request failed (prices affected obv but no crash of the whole app), fetchDmm could crash completely. but no arguing, if there is a better way - cool, lets use it.

> I'm stil debating internally on whether we want every rpc call in the api to benefit from this feature or not, wdyt?

retry looks obviously good, but isn't it already added on the transport level of viem?
its so much better if we dont need to wrap every request with this promise and it just retries under the hood.

btw im fully support to merge it early to avoid more and more conflicts, so proceed to merge if you like

seguido:
My understanding was that the retry of requests was only on transport layer errors, so rpc errors themselves weren't being retried (only http errors), I'll check the implementation to clarify on this

PR #1146, opened 2023-06-14
add new equilibre pools

schnarkus:


src/api/stats/getAmmPrices.ts
```diff
@@ -605,6 +605,7 @@ const seedPeggedPrices = {
   hUSDT: 'USDT', // HOP
   aWMATIC: 'MATIC', // Aave
   aWETH: 'ETH', // Aave
+  axlUSDT: 'axlUSDC', // Axelar
```
roman-monk, line 608: i dont think they are pegged, need to find volatile LP with axlUSDT

PR #1147, opened 2023-06-20
Add Chronos LPs

shatterproof:
- ARB-USDC.e vLP `chronos-arb-usdc.e` 6.1% (vs 6.0%)
- RDNT-ETH vLP `chronos-rdnt-weth` 9.8% (vs 3.2%)
- rETH-ETH vLP `chronos-weth-reth` 6.0% (vs 5.9%)
- MAI-LUSD sLP `chronos-mai-lusd` 12% (vs 3.5%)
- ETH-LQTY vLP `chronos-weth-lqty` 10% (vs 5.7%)

Also adds `arb:chronos:add` command, sets correct Beefy fees, switches to USDCe oracle ID, and fixes type bug (`bool` → `boolean`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "arb:swapfish:add": "ts-node scripts/add-farm.ts --network arbitrum --project swapfishArb --newFee true",
     "arb:arbidex:add": "ts-node scripts/add-farm.ts --network arbitrum --project arbidex --newFee true",
     "arb:ramses:add": "ts-node scripts/add-solidly.ts --network arbitrum --project ramses --newFee true --lp",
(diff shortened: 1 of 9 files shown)


roman-monk:
i think we need to filter these ones with the idea to not spam the app with the empty dead vaults.
2 simple filters:
1) 1M-ish TVL in farm
2) there is no same LP with higher APY

for instance frxETH LPs on ramses have better APY so nobody will deposit here.

additionally we have really bad capturing on chronos because staking directly could earn more over time.
so please be wise, most of these vaults will be empty bringing no fees.

PR #1148, opened 2023-06-20
Add Pearl MAI-USDR sLP

shatterproof:
- MAI-USDR sLP `pearl-mimatic-usdr`

Also includes `polygon:pearl:add` command and other platform data for Pearl.


roman-monk:
70k tvl

PR #1150, opened 2023-06-23
Add RDNT-BNB PCS Farm + RDNT Token for BSC

Bloufire:
Proposal to add an RDNT-BNB vault on PancakeSwap. 240k locked on PCS, 18M total liquidity in the pool (most of it locked on Radiant)


roman-monk:
pool on thena looks much better (x5 APY), not native BNB though

image

Bloufire:
To be honest, it was precisely my objective to offer this vault next. Being my first contribution I wanted to make sure it was done correctly with BNB/RDNT on PCS.

Regarding the pair on Thena, the fact that it is not native BNB goes a little against what Radiant recommends, which support native BNB for their liquidity mining rewards

PR #1164, opened 2023-07-03
Add veSync

shatterproof:


src/data/zksync/veSyncStableLpPools.json
```diff
@@ -0,0 +1,22 @@
+[
+  {
+    "name": "vesync-busd-usdc",
```
roman-monk, line 3: please drop it, too low tvl and its not gonna grow

src/api/stats/getNonAmmPrices.ts
```diff
@@ -90,6 +91,7 @@ export async function getNonAmmPrices(tokenPrices: Record<string, number>): Prom
   let breakdown = {};
 
   const promises = [
+    getveSyncStablePrices(tokenPrices),
```
roman-monk, line 94: after you drop busd-usdc this can be removed

src/api/stats/zksync/getveSyncApys.js
```diff
@@ -0,0 +1,48 @@
...
+    pools: pools,
+    oracleId: 'VSzk',
+    oracle: 'tokens',
+    decimals: getEDecimals(VS.decimals),
+    reward: VS.address,
+    boosted: false,
+    // log: true,
+  });
+
+  let apys = {};
```
roman-monk, line 29: everything below is not needed, it will be broken down on higher level.
please copy any other getSolidlyGaugeApy, i know you probably copied one, well that was bad example

PR #1184, opened 2023-07-27
Add new OOE pools

kexleyBeefy:
Reward pool has different signatures so the `getRewardPoolApys` template is adjusted to take extra params.


roman-monk:
maybe makes sense to make it as separate getApy file?
i think its clearly not RewardPool (need different strat as well), and its kinda unique with its function names

kexleyBeefy:
@roman-monk Created a unique APY calculation for OOE and removed the additions from the common template

PR #1189, opened 2023-07-31
Viem: chainlink + rate limiting

ReflectiveChimp:
`RPC_RATE_LIMIT` - default false, set to true to enable RPC rate limiting

There is a queue per rpc domain, default is 5 requests per second.

`RPC_RATE_LIMIT_CONCURRENCY`- default 2, max pending requests at same time
`RPC_RATE_LIMIT_INTERVAL_CAP`- default 5, max requests per interval
`RPC_RATE_LIMIT_INTERVAL`- default 1000, interval for above call


seguido:
Ty chimp, this will come in super handy

seguido (approved):

PR #1217, opened 2023-08-16
Fix BaseSwap APYs

kexleyBeefy:


src/api/stats/base/getBaseSwapApys.ts
```diff
@@ -0,0 +1,72 @@
...
+    const apy = yearlyRewardsInUsd.dividedBy(totalStakedInUsd);
+    apys.push(apy);
+  }
+
+  return apys;
+};
+
+const getMasterChefData = async () => {
+  const masterchefContract = fetchContract(masterchef, BaseSwapMasterChefAbi, chainId);
+  const blockRewards = new BigNumber((await masterchefContract.read.bswapPerSec()).toString());
```
roman-monk, line 50: @seguido could this affect performance? 1 await after another one without Promise.all, or is it fine?

PR #1238, opened 2023-09-02
Add validator performance to api

tilty-mctiltface:


src/api/validators/index.ts
```diff
@@ -0,0 +1,18 @@
+import { getValidatorPerformance, ValidatorPerformance } from './validators';
+
+const TIMEOUT = 5 * 60 * 1000;
+
+export const validatorPerformance = async ctx => {
+  try {
+    ctx.request.socket.setTimeout(TIMEOUT);
+    const validatorPerformance: ValidatorPerformance | undefined = await getValidatorPerformance();
```
seguido, line 8: since `getValidatorPerformance` is not an async function, there's no need to set a socket timeout nor to await the `getValidatorPerformance` call

src/router.js
```diff
@@ -17,13 +17,17 @@ const { bifibuyback } = require('./api/stats/bifibuyback/index');
...
+const { validatorPerformance } = require('./api/validators/index');
+
 const {
   vaultZapSupport,
   vaultZapSupportDebug,
   proxyOneInchSwap,
   proxyOneInchQuote,
 } = require('./api/zaps');
 
+router.get('/validator/performance', validatorPerformance);
```
seguido, line 29: It looks weird to me for us to support `/validator/performance` when we don't support `/validator`.

src/api/validators/validators.ts
```diff
@@ -0,0 +1,60 @@
...
+    validatorPerformance = {
+      eth: await getEthValidatorPerformance(),
+      ftm: await getFtmValidatorPerformance(),
+      fuse: await getFuseValidatorPerformance(),
+    };
+    console.log('> updated validator performance');
+    saveToRedis();
+  } catch (e) {
+    console.error('> validator performance initialization failed', e);
+  }
```
seguido, line 44: These 3 requests could be done concurrently so as to speed up the update. Right now `getFtmValidatorPerformance` doesn't start till `getEthValidatorPerformance` finishes.

This could look something like:
`const performances = await Promise.allSetled([getEth...(), getFtm...(), getFuse...()]);`
and assigning after this

src/utils/fetchEthValidatorData.ts
```diff
@@ -0,0 +1,35 @@
+import { ethers } from 'ethers';
```
seguido, line 1: Let's move this file out of utils and into `/api/validators`

src/utils/fetchFtmValidatorData.ts
```diff
@@ -0,0 +1,65 @@
+import BigNumber from 'bignumber.js';
```
seguido, line 1: let's move this file into /api/validators too

seguido (changes requested):

src/utils/fetchEthValidatorData.ts
```diff
@@ -0,0 +1,35 @@
...
+    }
+    return {
+      totalPerformanceEther: ethers.utils.formatUnits(
+        ethValidatorData.data[0].performancetotal.toString(),
+        'gwei'
+      ),
+    };
+  } catch (e) {
+    console.error('> fetchEthValidatorTotalPerformance', e);
+  }
```
seguido, line 34: If we return null or undefined when an error arrises, we'd be setting the local stored value to those values since this is the code calling this function:

```    
validatorPerformance = {
      eth: await getEthValidatorPerformance(),
      ftm: await getFtmValidatorPerformance(),
      fuse: await getFuseValidatorPerformance(),
    };
```

seguido: I'd suggest we throw an error when something happens and then work with the PromiseSettledResult from all the 3 performance calls, updating the stored value only for those calls that succeeded

seguido (changes requested):

src/router.js
tilty-mctiltface: What's a solution to this?  Or i can have it /validators, and can add support for /validators to return just the validators i guess

seguido: Yeah maybe `/validator` or `/validator-performance`

seguido (approved):

seguido (approved):

PR #1241, opened 2023-09-06
Feat/gov vaults endpoint

erenjaegar77:


src/api/stats/getMultichainVaults.js
```diff
@@ -15,9 +16,13 @@ const REFRESH_INTERVAL = 5 * 60 * 1000;
...
 
 let vaultsByChain = {};
+let govVaultsByChain = {};
 let multichainVaults = [];
+let multichainGovVaults = [];
 let multichainVaultsCounter = 0;
+let multichainGovVaultsCounter = 0;
 let multichainActiveVaultsCounter = 0;
+let multichainGovActiveVaultsCounter = 0;
 let vaultsByID = {};
```
seguido, line 26: I believe many of these fields are legacy ones that aren't needed. No need to have counters I'd say. It is useful to have the vaultsByID/govVaultsByID so I'd keep those. We could only keep `vaultsByChain, multichainVaults and vaultsByID` and their gov vault counterparts

src/api/stats/getMultichainVaults.js
```diff
@@ -32,6 +37,14 @@ export function getVaultByID(vaultId) {
   return vaultsByID[vaultId];
 }
 
+export function getMultichainGovVaults() {
+  return multichainGovVaults;
+}
+
+export function getSingleGovVaults(chain) {
```
seguido, line 44:
```suggestion
export function getSingleChainGovVaults(chain) {
```

src/api/stats/getMultichainVaults.js
```diff
@@ -46,13 +59,14 @@ async function updateMultichainVaults() {
...
       await saveToRedis();
     }
 
     console.log(
-      `> Vaults for ${fulfilled.length}/${
-        results.length
-      } chains updated: ${multichainVaultsCounter} vaults (${multichainActiveVaultsCounter} active) (${
+      `> Vaults for ${fulfilled.length}/${results.length} chains updated: ${
+        multichainVaultsCounter + multichainGovVaultsCounter
+      } vaults (${multichainActiveVaultsCounter + multichainGovActiveVaultsCounter} active) (${
```
seguido, line 69: We can infer the counters from the vault array, no need to keep/use them

src/api/stats/getMultichainVaults.js
```diff
@@ -81,18 +95,36 @@ function buildFromChains() {
   serviceEventBus.emit('vaults/updated');
 }
 
+function buildFromGovChains() {
+  multichainGovVaults = Object.values(govVaultsByChain).flat();
+  multichainGovVaultsCounter = multichainGovVaults.length;
+  multichainGovActiveVaultsCounter = sumBy(multichainGovVaults, vault =>
+    vault.status === 'active' ? 1 : 0
+  );
+  vaultsByID = keyBy(multichainGovVaults, 'id');
```
seguido, line 104: Careful, you're overwriting `vaultsByID` with the govVaults, this should be done on a separate variable

src/utils/getVaults.js
```diff
@@ -1,7 +1,6 @@
 const getVaults = async vaultsEndpoint => {
   try {
     let vaults = await fetch(vaultsEndpoint).then(res => res.json());
-    vaults = vaults.filter(vault => !vault.isGovVault);
     return vaults;
```
seguido, line 4: Did you check how this affects every usage of `getVaults`, this function is also used from `getChainTvl` and it'd be getting new objects that aren't vaults.

src/api/vaults/types.ts
```diff
@@ -30,4 +30,5 @@ export type Vault = {
   pricePerFullShare: BigNumber;
   createdAt: number;
   chain: ApiChain;
+  totalSupply?: number;
```
seguido, line 33: Gov vaults aren't the same as vaults, it might be worth having different types for each so it's easier to work with them on the api's side

seguido (changes requested):

src/api/stats/getChainTvl.js
```diff
@@ -81,10 +78,13 @@ const getGovernanceTvl = async (chainId, governancePool) => {
...
 
-  const excludedBalance = excludedBalances.reduce(
-    (tot, cur) => tot.plus(cur.times(tokenPrice).dividedBy(governancePool.tokenDecimals)),
-    new BigNumber(0)
-  );
+  let excludedTvl = 0;
+
+  if (govPool.excluded) {
+    console.log(govPool.excluded);
+    console.log(tvls[chainId][govPool.excluded]);
```
seguido, line 85: Please remove logs before pushing for review

src/api/stats/getChainTvl.js
```diff
@@ -8,12 +8,15 @@ const { default: BeefyVaultV6Abi } = require('../../abis/BeefyVault');
 
 const getChainTvl = async chain => {
   const chainId = chain.chainId;
-  const vaults = await getVaults(chain.vaultsEndpoint);
-  const vaultBalances = await getVaultBalances(chainId, vaults);
+
+  let vaults = await getVaults(chain.vaultsEndpoint);
+  const lpvaults = vaults.filter(vault => !vault.isGovVault);
```
seguido, line 13:
```suggestion
  const lpVaults = vaults.filter(vault => !vault.isGovVault);
```

src/api/stats/getChainTvl.js
```diff
@@ -1,19 +1,27 @@
...
-  const vaults = await getVaults(chain.vaultsEndpoint);
-  const vaultBalances = await getVaultBalances(chainId, vaults);
+
+  const lpVaults = getSingleChainVaults(apiChain);
+
+  const govVaults = getSingleChainGovVaults(apiChain);
+
+  const vaultBalances = await getVaultBalances(chainId, lpVaults);
+
+  const govVaultBalances = await getGovVaultBalances(chainId, govVaults);
```
seguido, line 20: What's with the spacing everywhere

seguido: vaultBalances and govVaultBalances can be fetched concurrently

src/api/stats/getChainTvl.js
```diff
@@ -38,11 +46,36 @@ const getChainTvl = async chain => {
...
+    }
+    const tvl = balance.times(tokenPrice);
+
+    let item = { [govVault.id]: 0 };
+
+    if (!tvl.isNaN()) {
+      item = { [govVault.id]: Number(tvl.minus(excludedTvl).toFixed(2)) };
+    }
+
+    tvls[chainId] = { ...tvls[chainId], ...item };
```
seguido, line 78: This block of code is super similar to the one above, it can be either merged with some conditions or refactored a bit

seguido (changes requested):

seguido (approved):

PR #1243, opened 2023-09-07
Add aerodrome pools & usdc native on op/base

erenjaegar77:
add Aerodrome vAMM-WETH-DAI
add Aerodome sAMM-USDC-USDbC
added USDC native on OP & Base


ReflectiveChimp:
Will any OP pools in API need updated to 'USDCe' instead of 'USDC'?

PR #1247, opened 2023-09-07
update usdce on arb/op for unique IDs

erenjaegar77:
Update usdc.e on optimism & arbitrum so each have isolated price


erenjaegar77:
This is to get different prices for USDCe on optimism and arbitrum

PR #1251, opened 2023-09-11
Add BVM axlUSDT-axlUSDC sLP

shatterproof:


roman-monk:
why `kaxl`? thats from kava. should be baxl or just axl?

PR #1258, opened 2023-10-04
Add support for multiple ls token aprs for aura apys

seguido:
Still missing cbETH's APR, I can't find where to get it from


seguido:
I ended up needing to add a factor to multiply the returned lsApr. Coinbase's API already returns the number as a percentage, we need to multiply it times 100 to normalize

seguido:
Please wait for weso's review to merge

PR #1270, opened 2023-10-25
Update: add new eth validators

erenjaegar77:


seguido:
You need to check how the responses are stored inside `multicallUtils`, the validator's address need to be set as the id to be in line with the proposed changes

seguido (approved):

PR #1304, opened 2023-11-20
Add Hop MAGIC & rETH

kexleyBeefy:


kexleyBeefy:
> seems oracles missing in ab

Oracle IDs added now

PR #1312, opened 2023-12-01
add aura arb ovn-wusd+ pool

schnarkus:


src/utils/fetchOVNPrices.ts
schnarkus, line 1: how about this?

MirthFutures: > how about this?

Close but we dont need to create a new file, you can just use this https://github.com/beefyfinance/beefy-api/blob/master/src/utils/fetchWrappedAaveTokenPrices.js.

PR #1315, opened 2023-12-06
Add Sonne Liquid Staking

kexleyBeefy:


MirthFutures:
@kexleyBeefy split up supply and vault apy? Put supply apy in trading apy slot.

PR #1329, opened 2023-12-28
add aero usdc-usd+

BustinJieber420:


src/data/base/aerodromeLpPools.json
```diff
@@ -1,4 +1,24 @@
...
+    "chainId": 8453,
+    "beefyFee": 0.095,
+    "lp0": {
+      "address": "0xB79DD08EA68A908A97220C76d19A6aA9cBDE4376",
+      "oracle": "tokens",
+      "oracleId": "USDC",
+      "decimals": "1e6"
+    },
+    "lp1": {
+      "address": "0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913",
```
MirthFutures, line 16: This is the USDC address?

src/data/base/aerodromeLpPools.json
```diff
@@ -1,4 +1,24 @@
 [
+  {
+    "name": "aerodrome-usdc-usd+",
+    "address": "0x418457Ca08fA5EC77f811B105F2c585cd051Ac10",
+    "gauge": "0x9701a079c6e80d91ce4c464c4a996237a27fe537",
+    "decimals": "1e18",
+    "chainId": 8453,
+    "beefyFee": 0.095,
+    "lp0": {
+      "address": "0xB79DD08EA68A908A97220C76d19A6aA9cBDE4376",
```
MirthFutures, line 10: This is USD+.

src/data/base/aerodromeLpPools.json
```diff
@@ -1,4 +1,24 @@
 [
+  {
+    "name": "aerodrome-usdc-usd+",
+    "address": "0x418457Ca08fA5EC77f811B105F2c585cd051Ac10",
```
MirthFutures, line 4: This is in the wrong file? This is stable pool and should be in aerodromeStableLpPools?

MirthFutures (commented):
There is a lot wrong with this. Before you submit the PRs it would be helpful to double check everything and make sure to run the api and make sure the results are what you expect.

src/data/base/aerodromeLpPools.json
BustinJieber420: fixed it all I hope. Thanks for your patience, shouldn't happen again as I can let the script do it now and even found out what "run the api" means!

PR #1358, opened 2024-02-15
Add Mantle token risks

kexleyBeefy:


packages/address-book/types/token.ts
```diff
@@ -10,7 +10,6 @@ interface Token {
   oracle?: 'tokens' | 'lps';
   bridge?: string;
   staked?: boolean;
-  //todo: add more risks
-  risks?: ReadonlyArray<'NO_TIMELOCK'>;
+  risks?: ReadonlyArray<string>;
```
erenjaegar77, line 13: i would do `ReadonlyArray<'NO_TIMELOCK'  | 'LARGE_HOLDERS'>` and then keep adding types to avoid having random issues

kexleyBeefy: Changed as advised

erenjaegar77 (approved):

PR #1401, opened 2024-04-04
Do not warn on old EOL mooTokenPrice fetch

ReflectiveChimp:
@kexleyBeefy how about this instead of #1397


ReflectiveChimp:
@kexleyBeefy that should already be covered by `...vault`?

kexleyBeefy:
@ReflectiveChimp I was copying in changes made by @seguido on main. I agree it would be covered, we can remove the line `type: vault.type,`

PR #1454, opened 2024-05-23
getApyBreakdown tidy

ReflectiveChimp:
Apply fee to clmApr
Remove clmApr from vaultApy
Return Number.MAX_SAFE_INTEGER instead of null for Infinity
Set values to undefined/remove key instead of 0 if data not available


seguido (approved):

ReflectiveChimp:
In https://github.com/beefyfinance/beefy-api/pull/1476 instead (outputting "Infinity" instead of max int)

PR #1467, opened 2024-06-12
Update/ Harvestable vaults endpoint by chain

erenjaegar77:
add support to
 `/harvestable-vaults/:chainId` 
 `/harvestable-vaults/id/:vaultId`
 `/cow-vaults/id/:vaultId`


DefiDebauchery:
If we can also replicate the `/id/:vaultId` path, that'd be a great inclusion for this PR, as well!

PR #1490, opened 2024-07-02
Move WETH-rETH-s pool from aerodromeStableLpPools to aerodromeLpPools

ying13:


kexleyBeefy:
@ying13 Why the change? The pool is in the correct file already?

ying13:
@kexleyBeefy I saw other pools in aerodromeStableLpPools were stable coins?

PR #1492, opened 2024-07-03
Add USDC-STAR stable pool on Base Aerodrome

ying13:
I want to add a USDC-STAR vault to app but the STAR token does not exist. What would be a better way to develop the vault without publishing updates to the address book?

### DIFF
--- a/packages/address-book/src/address-book/base/tokens/tokens.ts
+++ b/packages/address-book/src/address-book/base/tokens/tokens.ts
@@ -1190,4 +1190,19 @@ export const tokens = {
       'NORMUS is the governance token of the launchpad eNORMUSPUMP, a pump.fun fork on Base.',
     bridge: 'native',
   },
+  STAR: {
+    name: 'STAR',
+    symbol: 'STAR',
+    oracleId: 'STAR',
+    address: '0xC19669A405067927865B40Ea045a2baabbbe57f5',
+    chainId: 8453,
+    decimals: 18,
+    logoURI:
+      'https://raw.githubusercontent.com/SmolDapp/tokenAssets/main/tokens/8453/0xc19669a405067927865b40ea045a2baabbbe57f5/logo.svg',
+    website: 'https://www.preon.finance/',
+    documentation: 'https://docs.preon.finance/',
+    description:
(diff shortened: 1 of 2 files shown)


roman-monk:
star unsafu iirc

ying13:
@roman-monk Thanks for the review and help. Is the reason not enough liquidity or no timelock? I see the market cap and volume are small. I checked the token contracts but not certain if there’s migrate or upgrade functionality. Out of curiosity, what else we need to update to add a new token? I deployed the contracts, updated the pools and address book in beefy-api, vault data and images in beefy-v2, the ran both repositories, I saw the new token icon was shown in the localhost app, but the values like TVL, APY are 0. Was that because the address book was not published? On Jul 4, 2024, at 03:20, roman-monk ***@***.***> wrote:﻿
star unsafu iirc

kexleyBeefy:
@ying13 STAR's largest holder (61%) is the team's multi-sig, that needs to be timelocked or at least be controlled in some way. Also unlike most tokens it is a debt token that can be minted through various contracts using price feeds, all of which need to be secured. It is quite hard to prove secureness on these kind of tokens without working with their team to check everything.

For some untimelocked tokens we can use the `[NO_TIMELOCK]` flag. But that is reserved for tokens which have large marketcap, huge demand and the team is well-known. But even then they have to be checked throughly and have very few areas of vulnerability.

Did you change the URL in this [line](https://github.com/beefyfinance/beefy-v2/blob/539d97ae5bffae4d39adc9a9177d09d98de9abcb/src/features/data/apis/beefy/beefy-api.ts#L20) to `http://localhost:3000`? That's how to get it look at your local API instead of the prod one. The APY and TVL should appear but without the address book package update you shouldn't see the info about the token at the bottom or able to zap in to the vault.

ying13:
@kexleyBeefy Thank you very much for the explanation! Yes I changed the beefy-api API_URL to `http://localhost:3000` and the APY and TVL are 0 as below, and there's no zap feature. 
Screen Shot 2024-07-04 at 09 48 29

ying13:
I went over this process wanted to know how to verify and add new tokens. I will close this PR and many thanks to @kexleyBeefy @roman-monk for your time providing the helpful information 🙏

PR #1505, opened 2024-07-14
Add 6 Aerodrome CLM pools on base

ying13:


kexleyBeefy:
@ying13 Internally we're currently developing vaults to go over Aero and Velo CLMs, so they've mostly been deployed already but not yet released which is why I believe this PR hasn't been touched. You're welcome to deploy CLMs for the other platforms that don't currently require the reward pools like Uniswap and Camelot

PR #1516, opened 2024-07-22
Add uniswap clm WETH-wstETH on base

ying13:


kexleyBeefy:
@ying13 The position width of 50 is far too wide for this correlated pair, it should be around 15. For reference the width should follow this kind of rule:

| Fee | Width |
|--------|--------|
| 100 | 15 |
| 500 | 175 |
| 3000 | 60 |
| 10000 | 50 | 

Could you please redeploy with an updated width?

ying13:
@kexleyBeefy thank you. I just redeployed them and tested in UI.

PR #1520, opened 2024-07-24
add arbitrum pancake clm

ying13:


ying13:
There are no sub oracles for USD+ and USDT+ in beefyOracle, could you help set them?

PR #1555, opened 2024-09-26
Add additional token routes

DefiDebauchery:


src/api/tokens/index.ts
```diff
@@ -30,3 +36,28 @@ export const getChainTokens = ctx => {
...
+};
+
+export const getChainNatives = ctx => {
+  try {
+    const chainId = ctx.params.chainId;
+    const native = getTokenNative(chainId);
+    const wrapped = getTokenWrappedNative(chainId);
+
+    ctx.status = native || wrapped ? 200 : 404;
+    ctx.body = [native, wrapped];
```
seguido, line 58:
```suggestion
    ctx.body = { native, wrapped };
```
Can we return this as an object, rather than an array?

seguido: It'll make it easier for those consuming

seguido (approved):

PR #1578, opened 2024-11-20
Add mooAerodromeBaseWETH-JARVIS

iamjackgale:
Jack's first vault 🎉


roman-monk:
can use existing pool, will just need to rename old vault to aerodrome-weth-jarvis-eol

roman-monk:
ok i thought old one is eol already but its not, so need both

PR #1595, opened 2025-01-04
Update /support odos sonic

erenjaegar77:


seguido:
Odos patched their gas price bug, we can merge

PR #1598, opened 2025-01-13
Add missing fraxtal ab rp factory addy

prevostc:


kexleyBeefy:
@prevostc New address needs checksum. Also are there CLM vault & strategy factories also that aren't listed for Fraxtal, or is only the reward pool is deployed?

PR #1627, opened 2025-03-25
Add gnosis clm

prevostc:


prevostc:
@MirthFutures 

> Strategy Factory has the wrong owner, has vault owner instead of strategy owner. 
Incorrect? Owner is 0x09D19184F46A32213DF06b981122e06882B61309 which is the strategy owner set in [the gnosis addressbook](https://github.com/beefyfinance/beefy-api/blob/master/packages/address-book/src/address-book/gnosis/platforms/beefyfinance.ts#L7)

> The Reward Pool Factory doesnt have the reward pool implementation added and still has an eoa owner.
fixed

PR #1640, opened 2025-05-02
Fraxtal North Star changes

ReflectiveChimp:
Please bump address book in https://github.com/beefyfinance/beefy-v2/pull/2675 and deploy after this API PR is deployed.

Fraxtal:
- Rename FRAX name/symbol to frxUSD 
  - id/oracleId stays as FRAX so we don't lose historical data
- Rename wfrxETH name/symbol to frxETH
 - oracleId changed to frxETH (both existed previously)
- Rename FXS name/symbol to wFRAX
  - id/oracleId stays as FXS so we don't lose historical data
- Point WNATIVE to FXS (wFRAX)

Ethereum:
- add new frxUSD token (0xCAcd6fd266aF91b8AeD52aCCc382b4e165586E29)
  - oracleId is set to frxUSD but a) that already exists on Sonic, b) Legacy FRAX should be convertable to frxUSD 1:1, so perhaps we should only have one oracle, FRAX (which is the oracle of frxUSD now...)
- Update description of FRAX to indicate that it is now being called LFRAX/Legacy FRAX off chain
- Update description of FXS to indicate that is now being called FRAX off chain

Ethereum/Other chains:
- Haven't changed name or symbol, but maybe we should? Or have both some how like "frxUSD (p. FRAX)" and "FRAX (p. FXS)"

TLDR left ids and oracleIds as they are as much as possible, to keep historical data, however this results in a lot of confusion and will 100% lead to wrong oracleId being entered at some point in the future

@roman-monk help please

### DIFF
--- a/packages/address-book/src/address-book/ethereum/tokens/tokens.ts
+++ b/packages/address-book/src/address-book/ethereum/tokens/tokens.ts
@@ -196,16 +196,30 @@ export const tokens = {
   },
   FRAX: {
     chainId: 1,
-    address: '0x853d955aCEf822Db058eb8505911ED77F175b99e',
+    address: '0x853d955aCEf822Db058eb8505911ED77F175b99e', // LFRAX, convertable to frxUSD
     decimals: 18,
-    name: 'Frax',
-    symbol: 'FRAX',
+    name: 'Frax', // onchain: "Frax", offchain: "Legacy Frax Dollar"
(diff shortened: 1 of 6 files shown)


roman-monk:
> * Haven't changed name or symbol, but maybe we should? Or have both some how like "frxUSD (p. FRAX)" and "FRAX (p. FXS)"

other chains aren't updated yet, lets see how they approach it

PR #1777, opened 2026-04-29
[VAULT ]add morpho-v2-eth-kpk-usdc-prime-v2

erenjaegar77:


kexleyBeefy:
This pool already exists under the id `morpho-eth-kpk-usdc`. Can we just use that existing entry for the name insead in the app PR as well?

PR #1821, opened 2026-06-02
Add Base Pancake SOL-USDC CLM

BeefyPower:


BeefyPower:
Don't merge, need to redeploy with vault strategy strategyMerkl

BeefyPower:
pending confirmation from Frondo that this is what is expected for the Pancake/Base merkl campaign

PR #1833, opened 2026-06-18
Two aero clms

erenjaegar77:
`aerodrome-cow-base-kellyclaude-usdc-v2` & `aerodrome-cow-base-virtual-weth-v3`


roman-monk (changes requested):
StrategyRewardPool is deprecated, please use StrategyMerkl which is the same but with merkl

PR #1843, opened 2026-07-06
Add Merkl APR to Ethereum Balancer APYs

frondoto:


ReflectiveChimp (changes requested):
`merklApr` is for apr user has to claim (e.g. CLM Pools)
for rewards the vault compounds the merkl apr should be added to `vaultApr` (so its gets fee subtracted and compounded)

PR #1848, opened 2026-07-08
feat: add ERC-4626 token prices fetching functionality

erenjaegar77:
New ERC-4626 module to fetch token prices
```json
"stcUSD": 1.069530500381838,
```

### DIFF
--- a/src/utils/fetchErc4626TokenPrices.ts
+++ b/src/utils/fetchErc4626TokenPrices.ts
@@ -1,6 +1,5 @@
 import BigNumber from 'bignumber.js';
 import { parseAbi } from 'viem';
-import { ETH_CHAIN_ID } from '../constants';
 import { addressBook, ChainId, Token } from '../../packages/address-book/src/address-book';
 import { fetchContract } from '../api/rpc/client';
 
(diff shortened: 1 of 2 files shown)


src/utils/fetchErc4626TokenPrices.ts
```diff
@@ -0,0 +1,60 @@
...
+    return assetsPerShare.times(underlyingPrice).toNumber();
+  });
+};
+
+const fetchErc4626TokenPrices = async (tokenPrices: Record<string, number>): Promise<Record<string, number>> => {
+  const results = await Promise.all([getErc4626Prices(tokenPrices, tokens.ethereum, ETH_CHAIN_ID)]);
+  const vaults = Object.values(tokens).flat();
+  return results
+    .flat()
+    .reduce<Record<string, number>>((acc, price, i) => ((acc[vaults[i][1].oracleId] = price), acc), {});
```
ReflectiveChimp: Can you skip adding the `NaN` ones? Then it will be missing for anything that tries to use it rather than propagating the NaN

ReflectiveChimp (changes requested):

PR #1883, opened 2026-08-12
Add Robinhood stock CLMs

frondoto:


packages/address-book/src/address-book/robinhood/tokens/tokens.ts
```diff
@@ -60,7 +60,82 @@ export const tokens = {
...
+    documentation:
+      'https://cdn.robinhood.com/assets/robinhood/legal/rhj_final_terms_for_tokenised_debt_securities_linked_to_united_states_oil_fund.pdf',
+    bridge: 'native',
+    tags: ['STOCK'],
+  },
+  rhAAPL: {
+    name: 'Apple • Robinhood Token',
+    address: '0xaF3D76f1834A1d425780943C99Ea8A608f8a93f9',
+    symbol: 'AAPL',
+    oracleId: 'rhAAPL',
```
roman-monk: good chance to move 'rh' as a suffix same other stocks, AAPLon, AAPLx, AAPLb, AAPLc.

prefix is used to distinguish chains. 
so if `AAPLrh` ever goes to another chain, it can be `baseAAPLrh`.
same as `AAPLon` can be `rhAAPLon` and `arbAAPLon`

